PR #2, opened 2019-03-13
[BUG]fix no homepage in Podfile

exc3lsior:
when pod install in iOS, error occur and hint no homepage in PodSpec


bartolkaruza:
Thank you @exc3lsior!

PR #29, opened 2019-04-06
Add Standard ESLint Config

matt-oakes:
Converts the ESLint config to use the standard, as per https://github.com/react-native-community/meta/issues/9.


bartolkaruza:
Thanks @matt-oakes !

PR #37, opened 2019-04-09
Add description to package.json

FrikkieSnyman:
The description is needed, else `pod install` complains about an invalid summary field in the podspec


bartolkaruza:
Thank you!

PR #46, opened 2019-04-23
Fix #45 ordering of photos when groupTypes is All

### LINKED ISSUE #45: iOS: Sort order unexpected
When attempting to request all photos, photos are returned sorted by collection, instead of being a complete list. This makes it difficult to help users pick their most recent photos taken.
Example:
```
RNCCameraRoll.getPhotos({
  first: 20,
  assetType: 'Photos',
  groupTypes: 'All',
})
```
Will return the 20 latest photos as expected, in the simulator. However, on a real device this may or may not return the most recent 20 photos, as it'll sort the Collections by "end date" which will essentially show the album that was last modified, and show all photos from that album regardless of their date.

A javascript workaround is theoretically possible but it would require to fetch *all* photos (e.g. `first: 9999..`) and then sort by `timestamp`, which goes against the purpose of using cursors.

After tinkering with the source, it's clear that the issue arises from first grabbing collections, then assets. Regardless of the `sortDescriptor`, everything will be grouped by collection:
```
    PHFetchResult<PHAssetCollection *> *const assetCollectionFetchResult = [PHAssetCollection fetchAssetCollectionsWithType:collectionType subtype:collectionSubtype options:collectionFetchOptions];
    [assetCollectionFetchResult enumerateObjectsUsingBlock:^(PHAssetCollection * _Nonnull assetCollection, NSUInteger collectionIdx, BOOL * _Nonnull stopCollections) {
      // Enumerate assets within the collection
      PHFetchResult<PHAsset *> *const assetsFetchResult = [PHAsset fetchAssetsInAssetCollection:assetCollection options:assetFetchOptions];
```

**Proposed solution:**
I propose that we add a condition that if `groupTypes` is `"All"`, then we use [`fetchAssetsWithOptions`](https://developer.apple.com/documentation/photokit/phasset/1624783-fetchassetswithoptions?language=objc) instead of enumerating collections. No change required to `sortDescriptor`

scarlac:
By collecting assets (aka photos) directly if `groupTypes` is `"All"` then we fix the ordering issue where some people were seeing old photos first, e.g. a slew of panoramas as the first photos listed.

Disclaimer: My Objective-C is extremely rusty. Very likely that I'm hacking around too much but this seems to work on my device in our project and it avoids a big if/else statement with mostly duplicated code.

### DIFF
--- a/ios/RNCCameraRollManager.m
+++ b/ios/RNCCameraRollManager.m
@@ -215,96 +215,108 @@ RCT_EXPORT_METHOD(getPhotos:(NSDictionary *)params
   if (groupName != nil) {
     collectionFetchOptions.predicate = [NSPredicate predicateWithFormat:[NSString stringWithFormat:@"localizedTitle == '%@'", groupName]];
   }
+  
+  BOOL __block stopCollections_;
+  NSString __block *currentCollectionName;
 
   requestPhotoLibraryAccess(reject, ^{
-    PHFetchResult<PHAssetCollection *> *const assetCollectionFetchResult = [PHAssetCollection fetchAssetCollectionsWithType:collectionType subtype:collectionSubtype options:collectionFetchOptions];
-    [assetCollectionFetchResult enumerateObjectsUsingBlock:^(PHAssetCollection * _Nonnull assetCollection, NSUInteger collectionIdx, BOOL * _Nonnull stopCollections) {
-      // Enumerate assets within the collection
-      PHFetchResult<PHAsset *> *const assetsFetchResult = [PHAsset fetchAssetsInAssetCollection:assetCollection options:assetFetchOptions];
-
-      [assetsFetchResult enumerateObjectsUsingBlock:^(PHAsset * _Nonnull asset, NSUInteger assetIdx, BOOL * _Nonnull stopAssets) {
-        NSString *const uri = [NSString stringWithFormat:@"ph://%@", [asset localIdentifier]];
-        if (afterCursor && !foundAfter) {
-          if ([afterCursor isEqualToString:uri]) {
-            foundAfter = YES;
-          }
-          return; // skip until we get to the first one
+    void (^collectAsset)(PHAsset*, NSUInteger, BOOL*) = ^(PHAsset * _Nonnull asset, NSUInteger assetIdx, BOOL * _Nonnull stopAssets) {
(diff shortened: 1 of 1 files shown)


JoshuaGross:
At first glance, this seems reasonable to me. I don't particularly love it but it seems necessary because of how the new iOS library works, so it seems there's some inherent ugliness in any general solution :) I like it, thanks so much for doing this! I think this will stabilize things again 👍

ios/RNCCameraRollManager.m
```diff
@@ -215,96 +215,108 @@ static void RCTResolvePromise(RCTPromiseResolveBlock resolve,
...
+              @"speed": @(loc.speed), // speed in m/s
+            } : @{})
+          }
       }];
-    }];
+    };
+
+    if ([groupTypes isEqualToString:@"all"]) {
+      PHFetchResult <PHAsset *> *const assetFetchResult = [PHAsset fetchAssetsWithOptions: assetFetchOptions];
+      currentCollectionName = @"";
```
JoshuaGross: Should this be `@"all"` instead? Not sure, but setting `@"group_name"` to an empty string could be a problem?

JoshuaGross: The only reasonable option I can see is hardcoding "All Photos" to maintain backwards-compatibility with how CameraRoll used to work on iOS with ALAssetLibrary, and I'm not sure I love that either.

scarlac: I'm honestly not quite sure here. It's the equivalent to the album name. "All Photos" is possible, but I'm not sure what used to be emitted in that field in this situation.
I can change it to whatever you think is the best option here.

scarlac: I changed it to `"All Photos"`

JoshuaGross: It would have been "All Photos" in the past (though I'm not sure if this was a localized string or not -- but based on some previous investigation, I don't think it's localized).

bartolkaruza:
@scarlac I like this a lot, there is a clear bug in the previous logic and I'm really grateful you are taking the time to just fix this. I'm taking the time today to review the change in detail.

scarlac:
@bartolkaruza Tip: Hide whitespace changes, as most of the diff is just moved code.

PR #53, opened 2019-05-04
Feature/save to specific album

SimonErm:
This PR adds the option to saveToCameraRoll to save a picture to a specific album.
It replaces the "type" param with "options", which is optional and contains "type"(optional) and "album"(optional). Because it replaces an old parameter, it's a **breaking change.**

Furthermore it fixes getPhotos for Android by checking the platform before setting the default value for groupType, because otherwise it throws an exception with "groupTypes is not supported on Android" if groupTypes is not undefined. Maybe we should just remove this exception instead.

I tested the PR on iPhone Xr(Simulator) and OnePlus3T(Device)


typings/CameraRoll.d.ts
```diff
@@ -77,7 +81,7 @@ export interface CameraRollStatic {
   /**
    * Saves the photo or video to the camera roll or photo library.
    */
-  saveToCameraRoll: (tag: string, type?: 'photo' | 'video') => Promise<string>;
+  static saveToCameraRoll(tag: string, options?: SaveToCameraRollOptions): Promise<string>
```
SimonErm: i have to remove the static modifier

SimonErm: done

typings/CameraRoll.d.ts
```diff
@@ -77,7 +81,7 @@ export interface CameraRollStatic {
   /**
    * Saves the photo or video to the camera roll or photo library.
    */
-  saveToCameraRoll: (tag: string, type?: 'photo' | 'video') => Promise<string>;
+  saveToCameraRoll(tag: string, options?: SaveToCameraRollOptions): Promise<string>
```
bartolkaruza: I think you can rework this to become a non-breaking change by adding the additional options parameter. I agree with you that it is slightly less aesthetically pleasing but I want to avoid introducing a breaking change for this change, it will trip up a lot of people for little gain

js/CameraRoll.js
```diff
@@ -163,7 +165,7 @@ class CameraRoll {
     if (!params.assetType) {
       params.assetType = ASSET_TYPE_OPTIONS.All;
     }
-    if (!params.groupTypes) {
+    if (!params.groupTypes&&Platform.OS!=='android') {
```
bartolkaruza: formatting

js/CameraRoll.js
```diff
@@ -100,7 +100,10 @@ export type PhotoIdentifiersPage = {
     end_cursor?: string,
   },
 };
-
+export type SaveToCameraRollOptions = {
+  type?: 'photo' | 'video' | 'auto',
```
bartolkaruza: this new 'auto' type is useful, but shouldn't be in the same PR. It's much easier to assess impact and approve PR's when there is exactly one feature proposed and implemented.

ios/RNCCameraRollManager.m
```diff
@@ -98,49 +98,83 @@ static void requestPhotoLibraryAccess(RCTPromiseRejectBlock reject, PhotosAuthor
...
+          if (success){
+            PHFetchResult *collectionFetchResult = [PHAssetCollection fetchAssetCollectionsWithLocalIdentifiers:@[placeholder.localIdentifier]
+                                                  options:nil];
+            collection = collectionFetchResult.firstObject;
+            saveBlock();
+          }else{
+            reject(kErrorUnableToSave, nil, error);
+          }
+        }];
+      }else{
```
bartolkaruza: check formatting throughout :)

bartolkaruza:
@SimonErm thanks for this! I like the approach in general and the ability to extend the options without introducing future breaking changes. This is still possible without the current breaking change though, so would you be interested in reworking your implementation to match that?

js/CameraRoll.js
SimonErm: To be honest i just added auto to use default values instead of checking for undefined, so i didn't thought that as a feature 😆

SimonErm:
i messed up the rebase. I am going to open a new PR.

PR #54, opened 2019-05-06
Use the MimeType reported by the media Cursor instead of trying to guess the mime type from uri

FrikkieSnyman:
This prevents the crash when the device contains images with a `#`.

This fixes https://github.com/facebook/react-native/issues/24468.

The guessing of the content type was introduced in https://github.com/react-native-community/react-native-cameraroll/commit/da20713a64858401bfaf0067b6659905f165a764#diff-7f48a66bcd0a56ced1913161061c6e3bR434

@kesha-antonov is there a specific reason why you opted to make use of guessing the mimetype from the uri instead of using the value reported by the media Cursor?

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
@@ -43,7 +43,6 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.net.URLConnection;
 import java.nio.channels.FileChannel;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


lunaleaps:
@FrikkieSnyman coincidental timing! I had a fix for this internally https://github.com/facebook/react-native/pull/24721/files -- would be great to have you and @kesha-antonov to decide which approach would be best

FrikkieSnyman:
Thanks @lunaleaps. Is there any specific reason why you decided to keep with using the filename to determine what the mime type is?

To me, this seems a bit unreliable, as proven by this entire `#` issue. Also, we do already have a mime type reported by the Media Cursor, why would we try to determine it on our own?

bartolkaruza:
Hi there, I'm going to go ahead and decide that this PR is the right fix based on it being available here :)

Thanks for the effort @FrikkieSnyman and @lunaleaps !

lunaleaps:
@FrikkieSnyman good points! I'm in favor of your approach as well

kesha-antonov:
@FrikkieSnyman Hey! Thanks for the fix. I'm not Android pro so your approach maybe be better

mk-nickyang:
Hey there, thanks a lot for this fix. This might be little off topic, but just wondering how do you guys manage to display the images with `#`? It just displays as a blank square for me, and no error was generated.
```
<Image
    source={{ uri: 'file:///storage/emulated/0/Download/downlo#ad.jpeg' }}
    style={{ width: 200, height: 200 }}
    onError={e => console.log(e)}
/>
```

nohorjo:
> Hey there, thanks a lot for this fix. This might be little off topic, but just wondering how do you guys manage to display the images with `#`? It just displays as a blank square for me, and no error was generated.
> 
> ```
> <Image
>     source={{ uri: 'file:///storage/emulated/0/Download/downlo#ad.jpeg' }}
>     style={{ width: 200, height: 200 }}
>     onError={e => console.log(e)}
> />
> ```

Hi @mk-nickyang , in my JS I had to work around this with `uri.replace('#', '%23').replace('?', '%3F')`. It would be good if this could be handled natively

PR #57, opened 2019-05-13
Forced default groupType all fix

### LINKED ISSUE #56: ANDROID: GetPhotos doesn't work because of groupTypes
Hello,

The lib doesn't work on Android because GroupType default value is set here in JS https://github.com/react-native-community/react-native-cameraroll/blob/1e37b294269eeb7175a9d952619fb22c4d70821d/js/CameraRoll.js#L167
but in Android if there is the groupType params a exception is thrown 
https://github.com/react-native-community/react-native-cameraroll/blob/1e37b294269eeb7175a9d952619fb22c4d70821d/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java#L220

The bug appears when i upgrade from 1.0.3 to 1.0.4.

I can make a PR to resolve this issue if you want. That is the best solution? ignore the groupType in Android? or something else?

bartolkaruza:
This PR tries to balance the intention of throwing the exception when groupType is supplied to an Android app running Cameraroll and the fact that now a default is supplied which causes #56 .
fixes #56

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,10 @@ build/
 local.properties
 *.iml
 
+**/.settings
+**/.classpath
+**/.project
+
 # BUCK
 buck-out/
 \.buckd/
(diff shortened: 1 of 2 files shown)


bartolkaruza:
We should just ignore this input for Android and I'll update the docs to indicate it is iOS only.

PR #71, opened 2019-07-29
chore(lib): lock lodash version to mitigate security risk

bartolkaruza:


chrisbobbe:
What's the security risk?

PR #84, opened 2019-09-20
Feat/include exiftime

FrikkieSnyman:
# Summary

We needed the exif timestamp with photos. We could do it by iterating over the nodes and getting the exif for each, but this is slow!

I added an option to include the exif `DateTime` timestamp with `getPhotos`. This is only for android!

## Test Plan
Call `getPhotos` with `includeExifTimestamp` set to `true`.

Results:

Screenshot 2019-09-20 at 15 34 43

set `includeExifTimestamp` on the `getPhotos` params.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

## TODO

- [x] rename `exifTimestamp` to `exif_timestamp`

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ Returns a Promise which when resolved will be of the following shape:
       * `isStored`: {boolean}
       * `playableDuration`: {number}
     * `timestamp`: {number}
-    * `exifTimestamp?`: {number} // Android only. Optional because not all images are guaranteed to have an exif timestamp
+    * `exif_timestamp?`: {number} // Android only. Optional because not all images are guaranteed to have an exif timestamp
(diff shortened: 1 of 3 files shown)


idrissakhi:
Is this always needed ? if not, we can close this PR.

FrikkieSnyman:
> Is this always needed ? if not, we can close this PR.

No, that's why it's an option you can pass in.

This value is useful if you need to understand when the actual photo/image was created, instead of when it was just added to the device.

Arjan-Zuidema:
Good to have this behind a config option. Could you resolve the conflicts so this can be tested?

PR #55, opened 2019-05-09
Feature/save to specific album

SimonErm:
The rebased version of #53. I messed up rebase last time so i created a new remote branch and this PR.

PR #94, opened 2019-10-10
[Android] Use DCIM direcotry for saveToCameraRoll

FrikkieSnyman:
# Summary

Use DCIM dir for saveToCameraRoll. We noticed an increase in `External media storage directory not available` errors for our app. We noticed that the react-native-cameraroll in the current react-native project is using the DCIM directory, and was changed from the MOVIES and PICTURES dir quite a while ago. 

See here: https://github.com/facebook/react-native/blame/15ecb60d6deb96fcb7b0ef70faccd10594ededa3/ReactAndroid/src/main/java/com/facebook/react/modules/camera/CameraRollManager.java#L136

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅     |

## Checklist

- N/A I added the documentation in `README.md`
- N/A I updated the typed files (TS and Flow)
- N/A I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
@@ -125,14 +125,8 @@ public class CameraRollModule extends ReactContextBaseJavaModule {
       File source = new File(mUri.getPath());
       FileChannel input = null, output = null;
       try {
-        File environment;
-        if ("mov".equals(mOptions.getString("type"))) {
-          environment = Environment.getExternalStoragePublicDirectory(
-                  Environment.DIRECTORY_MOVIES);
-        } else {
-          environment = Environment.getExternalStoragePublicDirectory(
-                  Environment.DIRECTORY_PICTURES);
-        }
+        File environment = Environment.getExternalStoragePublicDirectory(
+          Environment.DIRECTORY_DCIM);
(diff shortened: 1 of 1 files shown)


bartolkaruza:
Thank you for this @FrikkieSnyman , it took me a while to find the time to investigate how this change happened in the first place. You are right that the current Facebook RN master branch uses the DCIM folder, much like yours. The problem on this repo was introduced by https://github.com/react-native-community/react-native-cameraroll/pull/55 which seems to have been rebased with a very old version of CameraRollManager.java.

bartolkaruza:
@FrikkieSnyman I think I concluded too quickly that the change back to Pictures/Videos was accidental. I suspect it is related to the ability of storing albums, @SimonErm can you confirm this?
I reverted the change for now, luckily I messed up the merge commit message, so no release was triggered. Once we figure this last bit out, we can merge the changes back to master.

FrikkieSnyman:
@bartolkaruza I'm guessing it's not included, right?

bartolkaruza:
@FrikkieSnyman you are right, I had to revert. The message above was triggered because both the merge and the revert were included in the 1.4.0 release.
> . I suspect it is related to the ability of storing albums
I can re-open these changes on top of the current master, which you can do as well if you want. We still need to check that storing albums is not broken.

PR #87, opened 2019-09-25
Fix warning for deprecated RCTBridge interface for RCTImageLoader

tomtargosz:
# Summary

Fixes #86

This library is using a class that got deprecated in React Native `0.61.0`. (https://github.com/facebook/react-native/blob/0.61-stable/React/CoreModules/RCTImageLoader.h)

This PR makes it so we're now using the recommended implementation.

### What's required for testing (prerequisites)?
Save an image to the camera roll and make sure no yellow box appears and that the image is successfully saved.

### What are the steps to reproduce (after prerequisites)?
In the current version of the library, save a photo to the camera roll and observe the yellow box.

Screen Shot 2019-09-25 at 2 13 12 PM

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    N/A     |

## Checklist

- [x] I have tested this on a device and a simulator
- [ ] I added the documentation in `README.md`
- [ ] I mentioned this change in `CHANGELOG.md`
- [ ] I updated the typed files (TS and Flow)
- [ ] I added a sample use of the API in the example project (`example/App.js`)

PR #109, opened 2019-11-12
Fix build error in RNCCameraRollManager.m

zcramos:
# Summary
After upgrading to 1.3.0 that includes #87, my builds started to fail on the step to link react-native-cameraroll with the following message:
`Undefined symbol: _OBJC_Class_$_RCTImageLoader`.
Screen Shot 2019-11-12 at 1 24 31 PM

I'm not sure what @tomtargosz's environment is like that allowed this to work, but it does not seem to work in 0.61.x with autolinked (cocoapods) dependencies. It looks like the recommendation made in that file by the react-native team is incorrect. I'm also not 100% confident this will work with RN < 0.60. I suspect the way `react-native-svg` fixed this may be backwards compatible - see [this commit](https://github.com/react-native-community/react-native-svg/commit/5452144468c2fbc157db7383e00ebd72bdad6af0) and [this one that specifically notes backwards-compatibility](https://github.com/react-native-community/react-native-svg/commit/3c22c9757277145a18a509c3aee36dddc6a610e7)

The correct way to fix this, which I see in [react-native-image-crop-picker](https://github.com/ivpusic/react-native-image-crop-picker/blob/master/ios/src/ImageCropPicker.m#L394) and in [several](https://github.com/facebook/react-native/blob/master/Libraries/Image/RCTImageEditingManager.mm#L60) [internal](https://github.com/facebook/react-native/blob/master/Libraries/Image/RCTImageView.m#L324) [react files](https://github.com/facebook/react-native/blob/master/Libraries/Image/RCTImageViewManager.m#L57), is to use `[self.bridge moduleForName:@"ImageLoader" lazilyLoadIfNecessary:YES]`.

## Test Plan

I made these changes in my local copy of the cocoapods and was able to compile.

A react-native app on RN 0.61.x that includes this package as an autolinked dependency via cocoapods. 

Try to build. It fails (for me - I haven't tried to repro with a fresh create-react-native-app but I don't see why it would work with a fresh app).

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |

## Checklist

- [x] ~I added the documentation in `README.md`~
- [x] ~I mentioned this change in `CHANGELOG.md`~
- [x] ~I updated the typed files (TS and Flow)~
- [x] ~I added a sample use of the API in the example project (`example/App.js`)~

### DIFF
--- a/ios/RNCCameraRollManager.m
+++ b/ios/RNCCameraRollManager.m
@@ -17,7 +17,7 @@
 
 #import <React/RCTBridge.h>
 #import <React/RCTConvert.h>
-#import <React/RCTImageLoader.h>
+#import <React/RCTImageLoaderProtocol.h>
 #import <React/RCTLog.h>
 #import <React/RCTUtils.h>
 
@@ -176,7 +176,7 @@ RCT_EXPORT_METHOD(saveToCameraRoll:(NSURLRequest *)request
       inputURI = request.URL;
       saveWithOptions();
     } else {
-      [[self->_bridge moduleForClass:[RCTImageLoader class]] loadImageWithURLRequest:request callback:^(NSError *error, UIImage *image) {
+      [[self.bridge moduleForName:@"ImageLoader" lazilyLoadIfNecessary:YES] loadImageWithURLRequest:request callback:^(NSError *error, UIImage *image) {
         if (error) {
           reject(kErrorUnableToLoad, nil, error);
           return;


bartolkaruza:
Thanks for this! Could you help me out and run a fresh react-native init on <0.60 app to see the effect of this change on older RN versions? Getting a confirmation, even if it's a failure, will make me more confident to merge this. Also we can make a note in the release notes and the readme about the loss of backwards compatibility.

bartolkaruza:
So I've asked @RSNara about the backwards compatibility and I'm going to see what he says first. I'm considering reverting https://github.com/react-native-community/react-native-cameraroll/pull/87 for the time being, which leaves a warning but at least doesn't cause any compile errors and is fully backwards compatible until a solution that would work in older versions of React Native can be found.

bartolkaruza:
So as far as I can tell, if we merge this in the compatibility will be:
cameraroll@<1.3.0  | react-native@<0.60
cameraroll@1.3.0 | react-native@0.60
(after merging the last PR) cameraroll@>1.3.0 | react-native@>0.60

I'll update the readme after releasing the fix. Thanks.

PR #120, opened 2020-01-03
Support filter media items between dates

r0b0t3d:
# Summary
Add `fromTime` and `toTime` to `getPhotos` to get media items in between

## Test Plan
Manual test on my device

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
@@ -127,8 +127,14 @@ public class CameraRollModule extends ReactContextBaseJavaModule {
       File source = new File(mUri.getPath());
       FileChannel input = null, output = null;
       try {
-        File environment = Environment.getExternalStoragePublicDirectory(
-          Environment.DIRECTORY_DCIM);
+        File environment;
+        if ("mov".equals(mOptions.getString("type"))) {
+          environment = Environment.getExternalStoragePublicDirectory(
+                  Environment.DIRECTORY_MOVIES);
+        } else {
+          environment = Environment.getExternalStoragePublicDirectory(
+                  Environment.DIRECTORY_PICTURES);
+        }
         File exportDir;
         if (!"".equals(mOptions.getString("album"))) {
           exportDir = new File(environment, mOptions.getString("album"));


bartolkaruza:
@thekevinbrown what's your view on this PR in light of your recent changes on Android (and your planned change for iOS pagination)?

android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
```diff
@@ -317,6 +324,20 @@ protected void doInBackgroundGuarded(Void... params) {
...
+        selectionArgs.add(mFromTime + "");
+      }
+      long toTime = mToTime;
+      if (!TextUtils.isEmpty(mAfter)) {
+        toTime = Long.parseLong(mAfter);
+      }
+      if (toTime > 0) {
+        selection.append(" AND " + SELECTION_DATE_TAKEN);
+        selectionArgs.add(toTime + "");
+      }
```
thekevinbrown: With the merging of #130, this can now be simplified to:

```java
      if (mFromTime > 0) {
        selection.append(" AND " + Images.Media.DATE_TAKEN + " > ?");
        selectionArgs.add(mFromTime + "");
      }
      if (mToTime > 0) {
        selection.append(" AND " + Images.Media.DATE_TAKEN + " <= ?");
        selectionArgs.add(mToTime + "");
      }
```
Which also has the side benefit of being able to now paginate through the results when they're filtered by time. Free wins are fun!

ios/RNCCameraRollManager.m
```diff
@@ -43,27 +43,44 @@ @implementation RCTConvert (PHAssetCollectionSubtype)
...
   NSString *const lowercase = [mediaType lowercaseString];
+  NSMutableArray *format = [NSMutableArray new];
+  NSMutableArray *arguments = [NSMutableArray new];
   
   if ([lowercase isEqualToString:@"photos"]) {
-    PHFetchOptions *const options = [PHFetchOptions new];
-    options.predicate = [NSPredicate predicateWithFormat:@"mediaType = %d", PHAssetMediaTypeImage];
-    return options;
+    [format addObject:@"mediaType = %d"];
+    [arguments addObject:[NSNumber numberWithInteger:PHAssetMediaTypeImage]];
```
thekevinbrown: Very minor thing, but I wonder if this can be simplified to:

```objectivec
[arguments addObject:@(PHAssetMediaTypeImage)];
```

ios/RNCCameraRollManager.m
```diff
@@ -43,27 +43,44 @@ @implementation RCTConvert (PHAssetCollectionSubtype)
...
   } else if ([lowercase isEqualToString:@"videos"]) {
-    PHFetchOptions *const options = [PHFetchOptions new];
-    options.predicate = [NSPredicate predicateWithFormat:@"mediaType = %d", PHAssetMediaTypeVideo];
-    return options;
+    [format addObject:@"mediaType = %d"];
+    [arguments addObject:[NSNumber numberWithInteger:PHAssetMediaTypeVideo]];
```
thekevinbrown: Same here:

```objectivec
[arguments addObject:@(PHAssetMediaTypeVideo)];
```
should be equivalent.

thekevinbrown:
@bartolkaruza This looks good, we also get the win that if there are 15 photos in the time range and you want a page size of 5, you can now paginate through the 15 using the `after` parameter, where before we fixed that bug we wouldn't have been able to do both together, e.g. using `after` would have clobbered the `fromTime` parameter.

On iOS this is fine, the change there will be pretty simple.
- Currently we're fetching a list of photos, then O(n) looping through the results, loading all the images until we find our `after` image, then grabbing a page.
- The new implementation will just fetch the list of photos, then have enough info in the `after` parameter to just pull the page out of that result set. Because these result sets are lazy evaluated, that should speed things up pretty dramatically. This PR changes the building of the predicate, but not what happens once there's a result set, so there's no overlap I can see.

bartolkaruza:
@r0b0t3d thank you for contributing! Could you have a look at the suggested changes? 

Thank you for the thorough review @thekevinbrown !

r0b0t3d:
@bartolkaruza Updated code follow by @thekevinbrown suggestion

bartolkaruza:
@r0b0t3d why did you close this PR? :)

r0b0t3d:
OMG, I forgot that I already open an PR on master. Because I open another PR so I restructure my branch  😬
Is there anyway to attach a branch to this PR or I need to open another?

bartolkaruza:
I think it's easiest to open a new PR from the branch where your work is. Does it only contain the specific changes from this PR?

r0b0t3d:
@bartolkaruza I opened #140 that contains only code from this PR. Please help to check.

PR #129, opened 2020-01-11
Show correct date in example code

thomaschaaf:
The timestamp is a unix timestamp to set a date from a unix timestamp it needs to be set in milliseconds.


bartolkaruza:
Thank a lot!

Issue #118, opened 2019-12-24
Pagination does not work correctly [Android]


Obi1Kennoby:
# Bug

I have around 1000 photos on my phone and when I try to load them by chunks (15 photos per time) only 109 photo loaded but if set chunk size to 200, I be able to load 325 photos

## Environment info
System:
    OS: macOS 10.15.1
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 32.89 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.12.0 - /usr/local/bin/node
    Yarn: 1.19.1 - /usr/local/bin/yarn
    npm: 6.12.0 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 26, 27, 28
      Build Tools: 27.0.3, 28.0.0, 28.0.2, 28.0.3
      System Images: android-26 | Google APIs Intel x86 Atom, android-26 | Google Play Intel x86 Atom, android-27 | Google APIs Intel x86 Atom, android-27 | Google Play Intel x86 Atom, android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom
      Android NDK: 14.1.3816874
  IDEs:
    Android Studio: 3.4 AI-183.6156.11.34.5522156
    Xcode: 11.2.1/11B500 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.5 => 0.61.5 
  npmGlobalPackages:
    react-native-git-upgrade: 0.2.7
  Device: Samsung Galaxy s10e, android 10

Library version: 1.3.0

Describe what you expected to happen:
It should be possible to load all available photos on a phone

## Reproducible sample code
```typescript
const IMAGES_PER_PAGE = 15

interface CurrentPageInfo {
  has_next_page: boolean // eslint-disable-line
  end_cursor?: string // eslint-disable-line
}

interface IImage {
  id?: string
  path: string
  mime: string
}

interface State {
  library: any[]
  results: IImage[]
}

export default class ImagePickerScreen extends Component<{}, State> {
  state: State = {
    library: [],
    results: [],
[...]

Issue #47, opened 2019-04-24
Load 20 images per time will lost some images


LuongTruong:
Hi everyone, I am using CameraRoll to implement the image Gallery and there is a bug

Bug details
========

I load 20 images first, after user scroll down to bottom, another 20 images load. This thing is: 
- I have total 201 images.
- If I load 1000000 images, I can get all 201 images.
- If I load 20 or 10 images per time, I only receive 190+ images.

To Reproduce
==========

I create a repo so that you guys can have a better investigation.

[Camera Roll Bug](https://github.com/truongluong1314520/RNCameraRollBug)

You can download or clone the repo, "npm install", and "react-native run-android"

Expected Behavior
==============

Lazy load or load all images, I can still get all of the images from the phone.

Environment
=========

React Native Environment Info:
    System:
      OS: macOS 10.14.1
      CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
      Memory: 2.64 GB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.15.0 - /usr/local/bin/node
      Yarn: 1.13.0 - /usr/local/bin/yarn
      npm: 6.9.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 16, 17, 18, 19, 21, 22, 23, 24, 25, 26, 27, 28
        Build Tools: 23.0.1, 25.0.2, 26.0.1, 26.0.2, 26.0.3, 27.0.3, 28.0.2, 28.0.3
        System Images: android-19 | Google APIs Intel x86 Atom, android-21 | Google APIs Intel x86 Atom, android-22 | Google APIs Intel x86 Atom, android-23 | Google APIs Intel x86 Atom, android-24 | Google Play Intel x86 Atom, android-25 | Google Play Intel x86 Atom, android-26 | Google Play Intel x86 Atom, android-27 | Google Play Intel x86 Atom, android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom
    IDEs:
      Android Studio: 3.1 AI-173.4907809
      Xcode: 10.1/10B61 - /usr/bin/xcodebuild
    npmPackages:
[...]

PR #130, opened 2020-01-20
Fix for pagination on Android.

thekevinbrown:
# Summary

In #47 and #118, users are experiencing problems with pagination on Android. Upon further investigation, I noticed that we're using Date Taken as the pagination value, each time selecting images older than the last image we cared about.

The problem is that there's no guarantee that Date Taken will be unique. In normal usage of a phone camera, it would be, but given that users can import photos from other sources, we were seeing data sets like the following:

```
2020-01-20 15:37:03.216 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000542458.jpg, Date Taken: 1566866253000
2020-01-20 15:37:03.217 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000553233.jpg, Date Taken: 1566866253000
2020-01-20 15:37:03.217 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000028462.jpg, Date Taken: 1566866252000
2020-01-20 15:37:03.217 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000040912.jpg, Date Taken: 1566866252000
2020-01-20 15:37:03.217 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000065562.jpg, Date Taken: 1566866252000
2020-01-20 15:37:03.218 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000083959.jpg, Date Taken: 1566866252000
2020-01-20 15:37:03.218 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000209400.jpg, Date Taken: 1566866252000
2020-01-20 15:37:03.218 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000258129.jpg, Date Taken: 1566866252000
2020-01-20 15:37:03.218 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000302227.jpg, Date Taken: 1566866252000
2020-01-20 15:37:03.219 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000381890.jpg, Date Taken: 1566866252000
2020-01-20 15:37:03.219 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000394523.jpg, Date Taken: 1566866252000
2020-01-20 15:37:03.219 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000407676.jpg, Date Taken: 1566866252000
2020-01-20 15:37:03.220 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000413603.jpg, Date Taken: 1566866252000
2020-01-20 15:37:03.220 27527-27703/com.camerarolllazyload I/CameraRoll: Image ID: file:///storage/emulated/0/Pictures/train2014/COCO_train2014_000000439940.jpg, Date Taken: 1566866252000
```

When using a page size of 3, all of those `1566866252000` images are stepped over in one page when they should not be.

The most deterministic way to handle this (in my opinion) was to keep the same sort order and just supply offset and limit parameters to the database.

In addition, we were previously abusing the Sort order parameter to pass a `LIMIT`. This happened to work, but could break pretty easily, so I refactored this to use the encoded query builder instead of passing it into the last parameter.

The only time this would be a breaking change is if people were doing something tricksy with the `after` parameter instead of just passing it back to us, but given that iOS works quite differently, and we were already putting numbers in there, I'd say the risk of this is pretty low.

It's also possible that someone will have more images than an `int` can handle, in which case we should be using a long. I did not test or plan for this scenario as it's outside of how my users use the module, but could if we want to.

## Test Plan

Using the reproduction case from #47 showed on my data set that we are now able to load all images, 3 at a time.

Easiest to do on the Android emulator with a data set that has duplicate Date Taken on the photographs. As you can see I'm using some images from the [COCO Dataset](http://cocodataset.org/#download).

The reproduction case in #47 does a pretty good job reproducing this.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [ ] I added the documentation in `README.md` (Not needed as this is a bug fix for an existing API)
- [ ] I mentioned this change in `CHANGELOG.md` (I'm unable to find this file)
- [ ] I updated the typed files (TS and Flow) (Not needed as this is a bug fix for an existing API)
- [ ] I added a sample use of the API in the example project (`example/App.js`) (Not needed as this is a bug fix for an existing API)

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
@@ -278,10 +278,6 @@ public class CameraRollModule extends ReactContextBaseJavaModule {
     protected void doInBackgroundGuarded(Void... params) {
       StringBuilder selection = new StringBuilder("1");
       List<String> selectionArgs = new ArrayList<>();
-      if (!TextUtils.isEmpty(mAfter)) {
-        selection.append(" AND " + SELECTION_DATE_TAKEN);
-        selectionArgs.add(mAfter);
-      }
       if (!TextUtils.isEmpty(mGroupName)) {
         selection.append(" AND " + SELECTION_BUCKET);
         selectionArgs.add(mGroupName);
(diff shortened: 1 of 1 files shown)


bartolkaruza:
Thank you @thekevinbrown , this is a long awaited fix, the community will be happy! 🎊

Issue #125, opened 2020-01-08
Exif data missing on iOS but working on Android


zubjorge:
# Bug

I am taking photos with **react-native-camera** and using this lib in combination with **piexifjs** and **react-native-fs** to set up Exif data. 

## Environment info

System:
    OS: macOS 10.15.2
    CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
    Memory: 2.00 GB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 13.2.0 - /usr/local/bin/node
    Yarn: 1.19.2 - /usr/local/bin/yarn
    npm: 6.13.1 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
    Android SDK:
      API Levels: 22, 23, 25, 26, 27, 28, 29
      Build Tools: 28.0.3, 29.0.2
      System Images: android-24 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5791312
    Xcode: 10.1/10B61 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6
    react-native: 0.60.4 => 0.60.4
  npmGlobalPackages:
    react-native-cli: 2.0.1

Library version: 1.2.1

## Steps To Reproduce

I'm setting EXIF data with **piexif**, creating the exif object and dumping it into the original photo data. Exif data is not present on iOS but it is on Android.

```
const exifObj = {
    '0th': zeroth,
    Exif: exif,
    GPS: gps,
  }
const exifStr = piexif.dump(exifObj)
const inserted = piexif.insert(exifStr, jpegData)
const newData = await removeJpegPrefix(inserted)
```
to finally save it with this library using **CameraRoll.saveToCameraRoll**. But when I go to the gallery, the photo has no exif data on iOS, but it has exif on Android.

Describe what you expected to happen:

1. iOS photo has Exif data.

Kuhne1:
The reason for this is that on the IOS side in the RNCCameraRollManager.m File.
The library uses, "creationRequestForAssetFromImage" if you look into the apple developer docs you will find this section. 

[...]

PR #134, opened 2020-01-21
changes to fix IOS MetaData

phyzical:
Pr of changes listed in https://github.com/react-native-community/react-native-cameraroll/issues/125

Allows for ios to save metadata and images

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

- [x ] I have tested this on a device and a simulator

### DIFF
--- a/ios/RNCCameraRollManager.m
+++ b/ios/RNCCameraRollManager.m
@@ -172,8 +172,8 @@ RCT_EXPORT_METHOD(saveToCameraRoll:(NSURLRequest *)request
   };
 
   void (^loadBlock)(void) = ^void() {
-      inputURI = request.URL;
-      saveWithOptions();
+    inputURI = request.URL;
+    saveWithOptions();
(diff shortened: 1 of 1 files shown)


phyzical:
as the developer listing the suggested fix said he is not a swift expert, neither am i. but i can confirm it fixed the metadata of the image.

ios/RNCCameraRollManager.m
```diff
@@ -172,20 +172,8 @@ static void requestPhotoLibraryAccess(RCTPromiseRejectBlock reject, PhotosAuthor
   };
 
   void (^loadBlock)(void) = ^void() {
-    if ([options[@"type"] isEqualToString:@"video"]) {
       inputURI = request.URL;
```
bartolkaruza: this needs one less indentation, because there is one less enclosing block

PR #156, opened 2020-03-23
Fix getPhotos crash if there is a quote in an album name

dmydry:
# Summary

This is a fix for cases when a user has albums with `'` in their names.

Simulator Screen Shot - iPhone 11 - 2020-03-23 at 12 50 44

1. Create an album on iOS device with `'` in its' name, for example `Men's Health`.
2. Try to use a `getPhotos` method.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

Simulator Screen Shot - iPhone 11 - 2020-03-23 at 12 49 42


dmydry:
I can fix the `ci/circleci: analyze` if you want to. But it doesn't relate to this fix

bartolkaruza:
> I can fix the `ci/circleci: analyze` if you'd like. But it doesn't relate to this fix

It's fixed on master, thanks though!

Issue #168, opened 2020-05-07
Thumbnail images


noobvim:
Hi, this package is unusable if someone wants to implement their own camera roll because the image sizes returned are huge, causing a very noticeable sluggish performance degradation when loading lots of images into FlatList. It would be great to have smaller thumbnail image sizes returned in addition to the regular full size images. Thanks!

react-native-image-resizer is not a valid workaround since it takes a long time to process images.

Similar to [https://github.com/react-native-community/react-native-cameraroll/issues/110](https://github.com/react-native-community/react-native-cameraroll/issues/110)

bartolkaruza:
Hey there, this is a useful addition to the library and I would be happy to take PR's solving the problem of resizing the images (or retrieving at a smaller size).

guy-shahine:
I created a PR from a solution that Jared Sumner provided but didn't have time to create a PR.  I looked in the repo for manual testing guidance but couldn't find any. Any pointers or someone jumping in to help testing this solution is appreciated.

dokkis:
I agree this is a must to have for such library and solution based on generating the thumbnail are not good for performance. I looked into it a little bit, for Android we could use:

https://developer.android.com/reference/android/provider/MediaStore.Images.Thumbnails.html 

It's deprecated starting from API Level 29 but it's available from API Level 1.
`Return thumbnail representing a specific image item. If a thumbnail doesn't exist, this method will block until it's generated. Callers are responsible for their own in-memory caching of returned values.`. Interesting thread on stackoverflow here: https://stackoverflow.com/questions/32782085/querying-the-mediastore-joining-thumbnails-and-images-on-id

For iOS I didn't investigate yet.

dokkis:
@bartolkaruza @noobvim I am trying to look into this improvement, I tried several method to generate or get existing thumbnail (only in Android for now):

Performance are observed on Xiaomi Mi A1.
1. ThumbnailUtils.extractThumbnail (around 70ms, this generated the thumbnail), could `always` return a base64 thumbnail
2. Images.Thumbnails.getThumbnail (around 10/20ms, should fetch an existing thumbnail), could return a base64 thumbnail (or null if is not yet generated)
3. MediaStore.Images.Thumbnails.queryMiniThumbnail (around 5/10ms, it fetches an existing thumbnail), could return the URI of an existing thumbnail (or null if is not yet generated)

[...]

Issue #138, opened 2020-01-30
How to make CameraRoll.getPhotos about 2x faster on iOS


Jarred-Sumner:
# Bug

I'm investigating why loading images from the camera roll within a custom image picker is slow. in my app.

There seem to be two main reasons:
1. Loading the actual images is slow *(not this issue)*
2. Fetching the list of images to load is slow **(this issue)**

From instruments profiling: (in release mode)
![image](https://user-images.githubusercontent.com/709451/73426158-75894e00-42e8-11ea-9465-fd7c33be8751.png)
![image](https://user-images.githubusercontent.com/709451/73426199-8df96880-42e8-11ea-8713-d949704d1729.png)

`[PHAssetResource assetResourcesForAsset]` is called here:  https://github.com/react-native-community/react-native-cameraroll/blob/24f141cada8c5174d4117a0177d228213520134f/ios/RNCCameraRollManager.m#L267-L288

What's happening here is:
- In order to get the mime type and filename for the `PHAsset`, we call `[PHAssetResource assetResourcesForAsset]`. 
- This function call is slow. Its not this library's fault that this function call is slow, it just is.

So....how do we fix this?

## Solutions

Solutions that come to mind:
1. Don't call `[PHAssetResource assetResourcesForAsset]` by default and omit `originalFilename` from the results. Only call that function if `[mimeTypes count]` > 0 or the developer passes in an argument saying they want to know the original filename.
2. Only call `[PHAssetResource assetResourcesForAsset]` if `[mimeTypes count]` > 0  and rely on getting the original filename via this:
```objc
NSString *const origFilename = [asset valueForKey:@"filename"];
```
The downside here is that this is an undocumented API. It seems to work on iOS 13, and Stack Overflow posts suggest it has worked since iOS 9.1. I will see if there are any issues with the app store. 

Locally, I opted for  solution 2 and `CameraRoll.getPhotos` is about 2x faster:
![image](https://user-images.githubusercontent.com/709451/73427251-34466d80-42eb-11ea-94da-29f3d608ea65.png)

Now the bottleneck is here, but I don't really think we can do anything about that:
![image](https://user-images.githubusercontent.com/709451/73427562-ebdb7f80-42eb-11ea-9a97-d5d87cdabc14.png)

[...]

Issue #110, opened 2019-11-16
Thumbnail images and performance recommendations?


billdami:
Hi,

Is there any way to request thumbnail sized versions of images to be returned by `getPhotos()` calls? I am noticing sluggish performance and high memory usage when attempting to render the returned images in a grid FlatList type UI using the core `Image` component (which does not appear to be choosing a smaller/lower resolution image on render, like it says it does in the docs, at least not on iOS).

I've seen some mentions of using something like `react-native-image-resizer` to create thumbnail versions on the fly, but it appears performance on this would also not be great. Is there any existing examples or solutions for performantly rendering cameraroll images in a large list like this? Ideally, using something like `FastImage` would be supported too, to keep loaded thumbnails in memory-cache, for quicker re-renders (i.e. scrolling images off the screen, then the scrolling them back into view).

Thanks!

bartolkaruza:
Hi there, thanks for reaching out. I'm going to close this issue as we're trying to keep this issue tracker focused on specific issues with the Cameraroll API. Your question to me looks like something that would get answered well on Stack Overflow.

billdami:
thanks for the reply @bartolkaruza. I did do some research on Stack Overflow and the like, and was not able to find a performant way to get smaller photos, while still using this library as the means to retrieve the list of photos themselves. 

As I am mainly concerned with iOS usage, I did notice this line in the iOS implementation:

https://github.com/react-native-community/react-native-cameraroll/blob/master/ios/RNCAssetsLibraryRequestHandler.m#L127

Which requests an image for the asset, using the dimensions of the original image itself. What im proposing is a way to customize these values as max height/width settings in the `getPhotos()` call. Being able to retrieve say, a 100x100 version of a 4K source photo seems like it would be very helpful when displaying grids of photos from the device's library in a performant/memory-conscious.

Is there a way to formalize this issue as a feature request? Or would it require an actual PR?

plus-:
Can we reopen this issue? @billdami proposition is valid, anyone trying to load the camera roll into thumbnails can't use the library as is at the moment.

pribeh:
I'll third @billdami 's recommendation. This seems like functionality that should reside in the library.

gordonwhitehouse:
Agree; facing same problem here.

noobvim:
> Agree; facing same problem here.

Just posted [https://github.com/react-native-community/react-native-cameraroll/issues/168](url). Hopefully they will re-engage with this problem.

[...]

PR #171, opened 2020-05-07
Improve get photos logic per Jared Sumner's solution

guy-shahine:
# Summary

We're facing slow performance speed when loading photos. Jared Sumner provided a solution that's explained in more details in #138 and other related issues are #110 and #168 

NOTE: CameraRollExample doesn't work in Xcode 11+ because it requires an upgrade of React Native.

## Test Plan

Since it's an improvement, I ran the existing tests and got no failures.

1. Run CameraRollExample project
2. Accept access to photo albums
3. Observe the images load speed

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    N/A     |

## Checklist

### DIFF
--- a/ios/RNCCameraRollManager.h
+++ b/ios/RNCCameraRollManager.h
@@ -13,8 +13,7 @@
 @interface RCTConvert (PHFetchOptions)
 
 + (PHFetchOptions *)PHFetchOptionsFromMediaType:(NSString *)mediaType
-                                       fromTime:(NSUInteger)fromTime
-                                         toTime:(NSUInteger)toTime;
+                                        subType:(PHAssetMediaSubtype)subType;
 
 @end
 
--- a/ios/RNCCameraRollManager.m
+++ b/ios/RNCCameraRollManager.m
@@ -17,6 +17,7 @@
 
 #import <React/RCTBridge.h>
 #import <React/RCTConvert.h>
+#import <React/RCTImageLoader.h>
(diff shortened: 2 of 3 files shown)


pribeh:
Quadruple pump this.

gordonwhitehouse:
Yes please. I quintuple pump this

pribeh:
@guy-shahine why did you close the pr?

guy-shahine:
@pribeh multiple reasons:
a) The package is still wired with an old structure and versions that prevents it from compiling with latest xcode
b) The tests passed but then I discovered a bug in Jared's code where a variable name was missing so tests are unreliable
c) I managed to get the example project to run in simulator but can't run on device (it requires an older iOS version that I don't have available)
d) No feedback, follow up or engagement from the maintainers

pribeh:
@guy-shahine thanks for sharing and your effort to implement this – I understand

tvpranav:
@guy-shahine Have we got any solution for thumbnail of images to make it more faster?

guy-shahine:
We couldn't solve it using this library so we changed our logic to use a different library that invokes the native photo picker.

aldimirprincipal:
> @tvpranav We couldn't solve it using this library so we changed our logic to use a different library that invokes the native photo picker.

Could you please tell me the library you are using? We are trying to implement a gallery but it's impossible on iOS since the memory leak and the slow performance

Issue #174, opened 2020-05-13
CameraRoll.getAlbums() broken in Android API Level 29


dokkis:
## Summary

CameraRoll.getAlbums() has 2 issues with API Level 29:
1. Exception thrown: `java.lang.IllegalArgumentException: Invalid column COUNT(*) as count.`
2. Exception get swallowed and doesn't return the promise here https://github.com/react-native-community/react-native-cameraroll/blob/master/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java#L430 so the function called CameraRoll.getAlbums() never get into the `then` or the `catch` part from client perspective

For the issue #1, it looks like GROUP BY support is removed from Android Q, that's why is not working.

see here:
* https://stackoverflow.com/questions/60623594/replacement-for-group-by-in-contentresolver-query-in-android-q-android-10-a
* https://stackoverflow.com/questions/468211/how-do-i-get-the-count-in-my-content-provider 

## Environment info

Android with API Level 29
Library version: 1.6.2

## Steps to reproduce

Try to use CameraRoll.getAlbums() on Android with API Level 29. No callback is never been called.

## Reproducible sample code

```
CameraRoll.getAlbums()
.then( (res) => console.log(res))
.catch( (err) => console.error(err))
```

No console statement is going to happen as there is an exception and has being swallowed without calling the reject callback.

dokkis:
neither GROUP BY or DISTINCT would work on Android Q, the only way that I can think of to achieve this behaviour is having a cursor to scroll all `MediaStore.Images.ImageColumns.BUCKET_DISPLAY_NAME` and found the set of this values, then for each of them create a cursor with filter that will return the size using `getCount`.

I don't know how would look like performance-wise.

r0b0t3d:
Just did some research. It seems the current solution doesn't works if you targeting api level 29 and above.
Found 2 keys in [ContentResolver](https://developer.android.com/reference/android/content/ContentResolver#QUERY_ARG_SQL_GROUP_BY) that are looks promising but they are added in Android R
```
QUERY_ARG_GROUP_COLUMNS
Specifies the list of columns (stored as a String[]) against which to group results.

QUERY_ARG_SQL_GROUP_BY
Key for an SQL style GROUP BY string that may be present in the query Bundle argument passed to ContentProvider#query(Uri, String[], Bundle, CancellationSignal).
```
[...]

PR #176, opened 2020-05-14
Fixed 'Invalid column COUNT(*) as count' on CameraRoll.getAlbums()

dokkis:
Fixed 'Invalid column COUNT(*) as count' on CameraRoll.getAlbums() due to GROUP BY support removed in Android Q.

# Summary

CameraRoll.getAlbums() has 2 issues:

1. Exception thrown: java.lang.IllegalArgumentException: Invalid column COUNT(*) as count with API Level 29.
2. Exception get swallowed and doesn't return the promise here https://github.com/react-native-community/react-native-cameraroll/blob/master/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java#L430 so the function called CameraRoll.getAlbums() never get into the `then` or the `catch` part from client perspective
For the issue #1, it looks like GROUP BY support is removed from Android Q, that's why is not working.

Related to: https://github.com/react-native-community/react-native-cameraroll/issues/174

Device/Emulator with API Level 29.

Use the getAlbums().

```
CameraRoll.getAlbums()
.then( (res) => console.log(res))
.catch( (err) => console.error(err))
```

This breaks in API Level 29 and the current PR support API Level 29 and below as well, it's using the same methodology without using `GROUP BY` that got removed in Android Q and having the group by logic by scrolling over all media using the cursor and count them.

I verified that the output of the method is the same by using `All`, `Photos` or `Videos` as assetType input compared to the old implementation. Both are ordered and both counts the same number of elements in the 3 above scenario in my device that has around 10k of photos.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [n/a] I added the documentation in `README.md`
- [n/a] I updated the typed files (TS and Flow)
- [n/a] I added a sample use of the API in the example project (`example/App.js`)

It's not a new functionality but a bug fix so no update on README or interface is required.

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
@@ -48,6 +48,8 @@ import java.io.IOException;
 import java.nio.channels.FileChannel;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
 
 import javax.annotation.Nullable;
 
@@ -374,14 +376,12 @@ public class CameraRollModule extends ReactContextBaseJavaModule {
     String assetType = params.hasKey("assetType") ? params.getString("assetType") : ASSET_TYPE_ALL;
     StringBuilder selection = new StringBuilder("1");
     List<String> selectionArgs = new ArrayList<>();
-    String bucketDisplayName = MediaStore.Images.ImageColumns.BUCKET_DISPLAY_NAME;
     if (assetType.equals(ASSET_TYPE_PHOTOS)) {
       selection.append(" AND " + MediaStore.Files.FileColumns.MEDIA_TYPE + " = "
               + MediaStore.Files.FileColumns.MEDIA_TYPE_IMAGE);
     } else if (assetType.equals(ASSET_TYPE_VIDEOS)) {
       selection.append(" AND " + MediaStore.Files.FileColumns.MEDIA_TYPE + " = "
               + MediaStore.Files.FileColumns.MEDIA_TYPE_VIDEO);
-      bucketDisplayName = MediaStore.Video.VideoColumns.BUCKET_DISPLAY_NAME;
     } else if (assetType.equals(ASSET_TYPE_ALL)) {
       selection.append(" AND " + MediaStore.Files.FileColumns.MEDIA_TYPE + " IN ("
               + MediaStore.Files.FileColumns.MEDIA_TYPE_VIDEO + ","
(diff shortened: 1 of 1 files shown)


dokkis:
@bartolkaruza can you kindly look into my PR? the release build also failed, but I am not sure if it's caused by my code, in the logs I see `AAPT2 aapt2-3.2.1-4818971-linux Daemon #0: Unexpectedly exit.`. I cannot retry the workflow to see if succeed. Thanks!

android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
```diff
@@ -413,22 +410,34 @@ public void getAlbums(final ReadableMap params, final Promise promise) {
         WritableArray response = new WritableNativeArray();
         try {
           if (media.moveToFirst()) {
+            Map<String, Integer> albums = new TreeMap<>();
```
r0b0t3d, line 413: What is the purpose of `TreeMap` here? If you just want to order album name, I think we do not need to do that. Basically, lib just return the album list, user may want to order the list by themself (by name or by count).

android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
```diff
@@ -413,22 +410,34 @@ public void getAlbums(final ReadableMap params, final Promise promise) {
...
+            }
           }
         } finally {
           media.close();
           promise.resolve(response);
         }
       }
-    } catch (Exception e) {}
-
+    } catch (Exception e) {
```
r0b0t3d, line 438: Could you consider to merge 2 try catch block into 1? the original implementation got problem here.

android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
dokkis: that was mainly for ordering, I can change to `HashMap` if we don't care about order and eventually clients can order on their and if they need to. Should I?

android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
dokkis: the best would be using the try-with-resources construct for the Cursor but is supported from API Level 19, If you do one main try you would need to have the `media` cursor reference outside of the try and check the nullability on the finally. I think I would rather keep as it is.

android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
r0b0t3d: It would be great!

bartolkaruza:
@dokkis the release build was just a glitch, I reran the workflow and it works fine. Thank you for the contribution!

Issue #83, opened 2019-09-18
[iOS] getPhotos take really long compared to previous package


simon-klikkie:
# Bug
I have recently moved to RN 0.60.5 from 0.59
Previously I was using the CameraRoll from `react-native` but with the upgrade to 0.60.6 I see that CameraRoll has been extracted and moved to this package.
Previously when I ran this
```
const cameraRollAssetsPromise = CameraRoll.getPhotos({
      assetType: 'Photos',
      first: 10000,
      groupTypes: 'All'
    });

const [cameraRollAssets] = await Promise.all([cameraRollAssetsPromise, albumAssetsPromise]);
```

This would return within a few seconds using the old CameraRoll.

But when using the CameraRoll from this package, this promise can take up to a minute to return.

Has this behaviour been experienced before? 
Is this just a limitation of the new library, or is this something that can be resolved?

## Environment info

React native info output:

```bash
System:
    OS: macOS 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 72.14 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.15.0 - ~/.nvm/versions/node/v11.15.0/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.7.0 - ~/.nvm/versions/node/v11.15.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5791312
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    @react-native-community/cli: ^2.9.0 => 2.9.0 
    react: ^16.9.0 => 16.9.0 
    react-native: ^0.60.5 => 0.60.5
```

Library version: 1.2.1

## Steps To Reproduce

1. Use CameraRoll.getPhotos with the old `react-native` package to get 10000 photos. results return in a few seconds.
2.Use CameraRoll.getPhotos with this package to get 10000 photos. results return in minutes.
...

Describe what you expected to happen:

1. I expected the response time of getPhotos to be similar

cporawat:
I have similar issue. 
[...]

PR #177, opened 2020-05-16
Updated example to build and run on latest React Native

hsource:
# Summary

While trying to submit a separate PR, I wanted to add something to the test app to allow testing that. I tried a few times, but due to bugs and changes since the version of React Native that was included, I couldn't get it to work.

This fix:

1. Upgrades `react-native` for the example app. It takes as much of the config from a fresh `npx react-native init` project as possible
2. Upgrades dependencies to be compatible with latest react-native
   - Made example app use Podfiles, per new versions
   - Upgraded `flow`, ran `yarn validate:flow`
   - Upgraded `detox`, ran build and test commands; made Android DetoxTest use androidx libraries and upgraded package.json commands
3. Adds new commands to package.json to run the example app, modeled off of `react-native-webview`
   - `yarn start:ios` runs iOS - tested on Simulator
   - `yarn start:android` runs Android - tested on real device

## Tips for reviewers

All manual changes to the React Native boilerplate have comments with "added for" in them, so a quick Ctrl + F wil find most of those.

## Test Plan

Tested by running:
```sh
yarn install
yarn test

yarn start:ios
yarn start:android

# Adjust settings for device in package.json if needed before these blocks
yarn test:detox:android:build:debug
yarn test:detox:android:test:debug
yarn test:detox:android:build:release
yarn test:detox:android:test:release
yarn test:detox:ios:build:debug
yarn test:detox:ios:test:debug
yarn test:detox:ios:build:release
yarn test:detox:ios:test:release
```

### What's required for testing (prerequisites)?

Nothing other than the commands above

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [X] I have tested this on a device and a simulator (no iOS device, only emulator)
- [ ] I added the documentation in `README.md`
- [ ] I updated the typed files (TS and Flow)
- [ ] I added a sample use of the API in the example project (`example/App.js`)

Issue #175, opened 2020-05-13
 After adding exif loading, the loading time of photos increased 10 times


jash90:
After adding this change https://github.com/react-native-community/react-native-cameraroll/commit/ade785e29d24228d3148a6f5581db7a885d029cc photo loading have been increased 10 times.
This feature should be optional.

Please undo the change or add the option to disable it.

bartolkaruza:
If you see the discussion in #178 I think the option to disable individual fields will solve this and other performance related issues.

ehsan6sha:
this is resolved?

PR #178, opened 2020-05-16
Added `include` parameter to getPhotos to let users tradeoff performance by omitting metadata

hsource:
# Summary

On a CameraRoll with ~30k photos, `getPhotos` was taking >3 minutes to get the metadata for all the photos. This was too slow for our application.

Digging through the ObjectiveC code, I saw that it was fetching detailed metadata for every photo, which took a lot of time. The metadata was only used to populate the filename and to filter by mimetype.

### Adding `include` to `getPhotos`

1. Adds an `include` param to `getPhotos`. Users can include `fileName`, `fileSize`, or `location`. The fields are null unless explicitly included
2. On iOS: Avoid calling `[PHAssetResource assetResourcesForAsset:asset]` in cases where it's not needed to dramatically improve performance
   - Most of the performance gain is here: this reduced run-times from 6 to 0.6 seconds
3. On iOS: Set `assetFetchOptions.fetchLimit = first;` when we know we won't be omitting any returned items
   - This only reduced running time by another 60%, down to ~0.2 seconds
4. On Android: Avoid calling `putLocationInfo` if the location is not needed, which reduces running time by 30%

With this fix, it takes 0.2 instead of 6 seconds to fetch 1k photos on iOS

### Making `toDate` consistently inclusive

- Made the logic on iOS for comparing `toDate` the same as that on Android by using `<=` instead of `<`

### Enabled Prettier for Typescript

- Changed the config to make it work for Typescript. It still isn't used for README.md, but a future quick commit can fix that

## Test Plan

Added automated test to verify the native module is tested correctly.

Also created an example app screen for testing.

1. Run the example app with `yarn start:ios`
2. Switch to the GetPhotosPerformanceExample.
3. Try tapping on both the buttons and see the speed difference. Ensure you have at least some photos on the phone. Here's a test with ~500 photos dragged onto it

| iOS | Android |
| --- | --- |
| iOS | Android |

Just `yarn install`, then `yarn start:ios` and `yarn start:android`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅ (unchanged behavior, but API is there)     |

## Checklist


ios/RNCCameraRollManager.m
```diff
@@ -70,7 +70,7 @@ + (PHFetchOptions *)PHFetchOptionsFromMediaType:(NSString *)mediaType
   }
   if (toTime > 0) {
     NSDate* toDate = [NSDate dateWithTimeIntervalSince1970:toTime/1000];
-    [format addObject:@"creationDate < %@"];
+    [format addObject:@"creationDate <= %@"];
```
pxpeterxu, line 73: Just a note: while this changes behavior some, this unifies the logic on Android (https://github.com/react-native-community/react-native-cameraroll/blob/master/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java#L331) and iOS, so is probably a net positive in the long run

js/__tests__/CameraRollTest.js
```diff
@@ -34,4 +34,9 @@ describe('CameraRoll', () => {
     await CameraRoll.getPhotos({first: 0});
     expect(NativeModule.getPhotos.mock.calls).toMatchSnapshot();
   });
+
+  it('Should call getPhotosFast', async () => {
```
pxpeterxu, line 38: Might be a bit more accurately named "should call NativeModules.getPhotos when calling getPhotosFast"

README.md
```diff
@@ -235,7 +236,38 @@ render() {
...
+**Parameters:**
+
+| Name   | Type   | Required | Description                                      |
+| ------ | ------ | -------- | ------------------------------------------------ |
+| params | object | Yes      | Expects a params with the shape described below. |
+
+All parameters have the same types and arguments as the parameters of the same names in [`getPhotos`](#getphotos).
+
+* `first` : {number}
+* `after` : {string}
```
pxpeterxu, line 262: It seems like `after` doesn't work with `getPhotosFast`, so we might want to update these docs too

js/CameraRoll.js
```diff
@@ -63,12 +57,47 @@ export type GetPhotosParams = {
...
+/**
+ * Params for the native `getPhotos` function, as implemented in the
+ * RNCCameraRoll module.
+ */
+export type GetPhotosNativeParams = GetPhotosParams & {
+  /**
+   * If provided, it's OK for the output to have empty filenames. This can
+   * improve performance on iOS when used by `getPhotosFast`.
+   */
+  allowEmptyFilenames?: boolean,
```
pxpeterxu, line 98: This might be better named as something else: e.g., "skipGettingFilenames"

hsource: Done!

README.md
hsource: Done!

js/__tests__/CameraRollTest.js
hsource: Renamed to `Should call native modules getPhotos when calling getPhotosFast`

ios/RNCCameraRollManager.m
hsource: Added a note in the initial commit message!

hsource:
@bartolkaruza - thanks for maintaining this package! Can you take a look at this change after #177

dokkis:
@hsource while you might make the implementation faster in iOS, that's not true for Android.

e.g. similar issue that you're trying to solve for Android: https://github.com/react-native-community/react-native-cameraroll/issues/175. In Android it's slow because we try to get location information from image one by one.

Instead of duplicating APIs, isn't a better solution providing `inclusion` settings to the `getPhotos` API, e.g. `includeFilenames`, `includeLocation`. This might allow several inclusion settings to extend to other fields to tune performance based on data that customers needs to retrieve.

A `getPhotosFast` implementation will result in an inconsistent implementation behaviour for iOS and Android, e.g. in iOS is fast if you don't return file names while in Android might be faster if you don't return location. This mean that the returned field has to be different for iOS and Android if you implement a separate API that is not good by design I believe.

Thoughts?

hsource:
> @hsource while you might make the implementation faster in iOS, that's not true for Android.
> 
> e.g. similar issue that you're trying to solve for Android: #175. In Android it's slow because we try to get location information from image one by one.
> 
> Instead of duplicating APIs, isn't a better solution providing `inclusion` settings to the `getPhotos` API, e.g. `includeFilenames`, `includeLocation`. This might allow several inclusion settings to extend to other fields to tune performance based on data that customers needs to retrieve.
> 
> A `getPhotosFast` implementation will result in an inconsistent implementation behaviour for iOS and Android, e.g. in iOS is fast if you don't return file names while in Android might be faster if you don't return location. This mean that the returned field has to be different for iOS and Android if you implement a separate API that is not good by design I believe.
> 
> Thoughts?

Seems like a good alternative API. I can try implementing this later!

dokkis:
out of curiosity, why do you need to get all items at once as opposed to fetch items using pagination?

hsource:
> out of curiosity, why do you need to get all items at once as opposed to fetch items using pagination?

We're trying to show places the user has traveled to on a map based on their photos, and the aggregate information is helpful for that!

dokkis:
> > out of curiosity, why do you need to get all items at once as opposed to fetch items using pagination?
> 
> We're trying to show places the user has traveled to on a map based on their photos, and the aggregate information is helpful for that!

The challenge then came with the Android implementation, https://developer.android.com/reference/android/provider/MediaStore.Video.VideoColumns#LONGITUDE
In API Level 29 they don't index those information anymore for privacy reasons, so there is no way to get this performant, related to this issue I believe: https://github.com/react-native-community/react-native-cameraroll/issues/175

bartolkaruza:
@hsource this is a valid case and it looks like a good implementation at the native level to me. I have similar concerns about the JS side API as @dokkis 

> Instead of duplicating APIs, isn't a better solution providing inclusion settings to the getPhotos API, e.g. includeFilenames, includeLocation. This might allow several inclusion settings to extend to other fields to tune performance based on data that customers needs to retrieve.

This suggestion looks good to me. I'm happy to review it again in that form. Adding another `getPhotos` will just confuse first time users (Why is there a fast and why would you want to use the slow one? And why is the fast one still slow on Android? ). By specifying inclusions, your iOS performance issue and the Android issue #175 can be resolved neatly without overly complicating the API. The attribute documentation for those inclusions is also the perfect place to document these performance concerns. A small header in the docs named performance, could add a reference to these attributes for those investigating their seemingly slow performing cameraroll implementation.

Perhaps you could implement the above inclusion mechanism for the filenames for iOS and you or someone else can add the Android fix using the same mechanism in a follow up PR?

bartolkaruza:
just for reference, this includes mechanism can also apply to https://github.com/react-native-community/react-native-cameraroll/pull/84

bartolkaruza:
I think this would be a fix for https://github.com/react-native-community/react-native-cameraroll/issues/83 as well.

hsource:
I changed the API and updated the documentation. Take a look - it should currently has options for `fileSize`, `filename`, and `location`.

r0b0t3d:
Not sure `include` will be a good param. It is breaking changes for current apps that need `fileSize`, `filename`, or `location`. Why don't we take `exclude` instead? For folks that don't need those params, they just exclude it. And it stills works well with current app.

dokkis:
@r0b0t3d I don't have a strong opinion in `include` vs `exclude` but I strongly agree on not altering the existing behaviour. `include` could either semantically include `*` everything,  if it's not provided or might be implemented with an `exclude` parameter.

js/CameraRoll.js
```diff
@@ -186,27 +205,37 @@ class CameraRoll {
...
-      params.groupTypes = GROUP_TYPES_OPTIONS.All;
-    }
     if (arguments.length > 1) {
       console.warn(
         'CameraRoll.getPhotos(tag, success, error) is deprecated.  Use the returned Promise instead',
       );
       let successCallback = arguments[1];
       const errorCallback = arguments[2] || (() => {});
-      RNCCameraRoll.getPhotos(params).then(successCallback, errorCallback);
+      const promise = RNCCameraRoll.getPhotos(params);
```
bartolkaruza, line 233: `getParamsWithDefaults` is missing here

android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
```diff
@@ -469,14 +475,18 @@ private static void putEdges(
     int sizeIndex = media.getColumnIndex(MediaStore.MediaColumns.SIZE);
     int dataIndex = media.getColumnIndex(MediaStore.MediaColumns.DATA);
 
+    boolean shouldGetLocationInfo = includeArrayContains(includeArray, "location");
```
bartolkaruza, line 478: As a policy, should we make the includes consistent between platforms? I understand that `filename` and `fileSize` have less performance impact on Android, but I think that leaving it on always will lead to issues like:
> There is no filename on iOS, Android works fine!
Consistently turning both off, gives people full control. By doing a platform check they can modify the includes array to tune it as needed.

bartolkaruza:
> Not sure include will be a good param. It is breaking changes for current apps that need fileSize, filename, or location. Why don't we take exclude instead? For folks that don't need those params, they just exclude it. And it stills works well with current app.

As a counter point, I think it would be better to make this a breaking change and to not include any of these fields by default. It will lead to leaner and faster apps for those that don't need these fields. The breaking change mechanism, through a major release, should protect those that rely on this information in the current library version. Most of the information fields are fairly recent additions and I think a minority of library users rely on the fields currently, which gives us an opportunity to make the library drastically faster for the majority. How do you feel about this point of view?

js/CameraRoll.js
hsource: Whoops! Good catch - I added it back

js/CameraRoll.js
```diff
@@ -186,27 +205,37 @@ class CameraRoll {
...
-    if (!params.groupTypes && Platform.OS !== 'android') {
-      params.groupTypes = GROUP_TYPES_OPTIONS.All;
-    }
     if (arguments.length > 1) {
       console.warn(
         'CameraRoll.getPhotos(tag, success, error) is deprecated.  Use the returned Promise instead',
       );
       let successCallback = arguments[1];
       const errorCallback = arguments[2] || (() => {});
-      RNCCameraRoll.getPhotos(params).then(successCallback, errorCallback);
```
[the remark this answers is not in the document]
hsource: This is actually correct I think. In the previous version of the code, it makes this call. And then, since there isn't an early return, it then makes a duplicate call and returns the promise anyways. This code maintains the same behaviour and return value as before, but only calls the native getPhotos once.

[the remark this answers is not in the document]
bartolkaruza: Ah you are right that it falls through to the second call. I guess it doesn't hurt to return the promise. Alternatively you could return without a value, but I'll leave that up to you.

android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
hsource: I agree. I did this, and updated the gifs in the description of the change to show this.

hsource:
> > Not sure include will be a good param. It is breaking changes for current apps that need fileSize, filename, or location. Why don't we take exclude instead? For folks that don't need those params, they just exclude it. And it stills works well with current app.
> 
> As a counter point, I think it would be better to make this a breaking change and to not include any of these fields by default. It will lead to leaner and faster apps for those that don't need these fields. The breaking change mechanism, through a major release, should protect those that rely on this information in the current library version. Most of the information fields are fairly recent additions and I think a minority of library users rely on the fields currently, which gives us an opportunity to make the library drastically faster for the majority. How do you feel about this point of view?

I also agree with making this a breaking change, and making the library fast-by-default, and return detailed metadata only on request.

By making things excluded by default and requiring an explicit include:

- Users who care about getting specific metadata when developing their app will see that it's not there, and add the parameters
- Users who don't care will just never get it, and get fast performance

If we instead made data included by default:

- Users who care about getting specific metadata would get all of it
- Users who don't care about metadata get slow performance, and only realize this when their app is in production and running on larger datasets

hsource:
@bartolkaruza Before pushing out this breaking change, I also think we should add `playableDuration`, and even `imageSize` (for width and height) to the `include` array. These operations both can take a good amount of time on Android.

hsource:
I'm done making all changes! If you'd like to push a breaking version change, I think it might be a good idea to also bundle in #187 too

bartolkaruza:
> I'm done making all changes! If you'd like to push a breaking version change, I think it might be a good idea to also bundle in #187 too

Really great work! I'll wait for one more review from anyone, then I'll merge this. With regards to the breaking changes, the cost of doing 2 major releases is very low with semantic-release. We can do them right after each other.

README.md
```diff
@@ -162,9 +162,13 @@ Returns a Promise with photo identifier objects from the local camera roll of th
   * `All`
   * `Videos`
   * `Photos` // default
-* `mimeTypes` : {Array} : Filter by mimetype (e.g. image/jpeg).
+* `mimeTypes` : {Array} : Filter by mimetype (e.g. image/jpeg). Note that using this will reduce performance slightly on iOS.
 * `fromTime` : {timestamp} : Filter from date added.
 * `toTime` : {timestamp} : Filter to date added.
```
pxpeterxu, line 167: Minor, but might be good to add the note about `fromTime` being exclusive and `toTime` inclusive here too!

hsource: Done making the change. Also mentioned that `fromTime` and `toTime` take milliseconds, while the returned `timestamp` is in seconds.

ios/RNCCameraRollManager.m
```diff
@@ -286,34 +296,47 @@ static void RCTResolvePromise(RCTPromiseResolveBlock resolve,
...
+            }
+          }];
 
-        if (!mimeTypeFound) {
-          return;
+          if (!mimeTypeFound) {
+            return;
+          }
         }
       }
```
r0b0t3d, line 342: Can't comment on the exact line. So I put it here.
You should move this code block to the end of function. Otherwise, it can't be executed because the last element is not added to list.
```
      // If we've accumulated enough results to resolve a single promise
      if (first == assets.count) {
        *stopAssets = YES;
        stopCollections_ = YES;
        hasNextPage = YES;
        RCTAssert(resolvedPromise == NO, @"Resolved the promise before we finished processing the results.");
        RCTResolvePromise(resolve, assets, hasNextPage);
        resolvedPromise = YES;
        return;
      }
```

bartolkaruza:
This will hopefully fix #83 and #14

ios/RNCCameraRollManager.m
hsource: This code was there before this change, and is working correctly (both before and after the change).

If we've already gotten the number the user has requested, we should not add any more items.

Keeping the code here instead at the end of the callback block also ensures that it runs correctly and returns 0 items if `first` is set to be 0.

hsource:
> @hsource Can you have a look at @r0b0t3d 's last comment? I'm comfortable merging and releasing this after that change and a retest.

I think the code should not be changed and can be merged as-is!

r0b0t3d:
@hsource Because you just add `assetFetchOptions.fetchLimit = first;`.
So let say first = 10, With the last item, `asset.count` has value 9.
So, if put that code block before adding to list, it will never be executed.
It is easy to test by adding `first`. You could try yourself.

hsource:
> @hsource Because you just add `assetFetchOptions.fetchLimit = first;`.
> So let say first = 10, With the last item, `asset.count` has value 9.
> So, if put that code block before adding to list, it will never be executed.
> It is easy to test by adding `first`. You could try yourself.

It works in that case still, and I've tested it.

Let's say the user puts `first: 10`.

1. `assetFetchOptions.fetchLimit` is set to 10
2. `[assetsFetchResult enumerateObjectsUsingBlock:collectAsset];` is run with the collectAsset block
3. The `collectAsset` block is run 10 times. On the 10th iteration
  - `if (first == assets.count) {` is false, since there are only 9 elements in the `assets` array
  - As such, we do not return, and add the 10th element to the `assets` array
4. The code continues executing, and we resolve the promise with the `assets` array, which now has 10 elements. The code that does this is at the end of the function

```mm
    if (!resolvedPromise) {
      hasNextPage = NO;
      RCTResolvePromise(resolve, assets, hasNextPage);
      resolvedPromise = YES;
    }
```

Let me know if you find something different while testing, and feel free to add a test case to the example app!

r0b0t3d:
That is the problem with load more function, user just want to load 10 items each time. So with current implementation, `hasNextPage` will be false if `first` is set.

hsource:
Ah, I see! Yeah - in that case, it sounds like the right solution would be to set `assetFetchOptions.fetchLimit` to `first + 1`. In that case, `hasNextPage` will be correct no matter what. Does that sound right to you?

r0b0t3d:
It will be great! Please go ahead.

hsource:
Actually, @r0b0t3d, I tested it and the previous implementation worked also.

Even if there were exactly 10 items, it would still say that `hasNextPage` is true.

| Before | After |
| --- | --- |
| image | image |

r0b0t3d:
@hsource Not sure that I get it. Could you make it clear with `Before` and `After`?
As I can see `has_next_page` is false in case `Before`

hsource:
Whoops, sorry for the confusion. I posted the wrong comment but the right screenshot. I submitted a new one with the fix you suggested! I also checked it.

| Before | After |
| --- | --- |
| image | image |

@bartolkaruza - this should be good to submit now. @r0b0t3d - sorry about the confusion

bartolkaruza:
Ok, I've made a mistake merging this in 1.8.0 as a minor release, it should have been a major release. I'm going to try to fix that issue manually.

bartolkaruza:
Ok, it's sorted now. 1.8.0 has this PR merged, 1.8.1 reverted and 2.0.0 included again.

PR #185, opened 2020-05-22
Use DCIM folder when album is not provided on Android.

dokkis:
# Summary

Adjusted README and implemented `save` function to Android to use `DCIM` if no album is provided.

Problem 1:
First of all there is a bug in the current handling of type param for Android in the save function. From doc it's either video or photo, but on implementation https://github.com/react-native-community/react-native-cameraroll/blob/master/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java#L133 it's checking over mov value. If I try to pass mov as value from JS, I got an error on the enum check. So there is no way to reach the current Environment.DIRECTORY_MOVIES dir in Android.

Problem 2:
If I try to simply replace Environment.DIRECTORY_PICTURES with Environment.DIRECTORY_DCIM, the save photo works (no errors) but I never see the new album appearing in the Photos app. I see the new album if I use Environment.DIRECTORY_PICTURES instead.

Problem 3:
Not sure if this is really a problem, but if you save without specifying an album, your media will go into Pictures "album" not into Camera, as opposed if you use DCIM that makes your media appearing to Camera "album/folder". This is a regression compared to the old saveToCameraRoll function.

android device/emulator

try to save to camera roll:
1. by providing an album
 a. with video type
 b. with image type
2. by not providing an album

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
@@ -129,16 +129,22 @@ public class CameraRollModule extends ReactContextBaseJavaModule {
       File source = new File(mUri.getPath());
       FileChannel input = null, output = null;
       try {
-        File environment;
-        if ("mov".equals(mOptions.getString("type"))) {
-          environment = Environment.getExternalStoragePublicDirectory(
-                  Environment.DIRECTORY_MOVIES);
+        boolean isAlbumPresent = !"".equals(mOptions.getString("album"));
+        
+        final File environment;
+        // Media is not saved into an album when using Environment.DIRECTORY_DCIM.
+        if (isAlbumPresent) {
+          if ("video".equals(mOptions.getString("type"))) {
+            environment = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_MOVIES);
+          } else {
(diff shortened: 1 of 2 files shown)


dokkis:
@bartolkaruza let me know if this makes sense to you :) thanks!

android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
```diff
@@ -129,16 +129,25 @@ protected void doInBackgroundGuarded(Void... params) {
...
-        File environment;
-        if ("mov".equals(mOptions.getString("type"))) {
-          environment = Environment.getExternalStoragePublicDirectory(
-                  Environment.DIRECTORY_MOVIES);
+        boolean isAlbumPresent = !"".equals(mOptions.getString("album"));
+        
+        final File environment;
+        // Media is not saved into an album when using Environment.DIRECTORY_DCIM.
+        if (isAlbumPresent) {
+            if ("video".equals(mOptions.getString("type"))) {
```
bartolkaruza, line 137: The formatting for Java files is 2 spaces in the rest of the code base

dokkis: @bartolkaruza should be fixed now, I didn't noticed it!

PR #187, opened 2020-05-23
Added imageSize and playableDuration to `include` param, deletes isStored

hsource:
# Summary

**This is another change that is not backwards compatible**

I realized that the Android implementation still had poor performance since it looped through individual images to populate `image.width/height` (some of the time) and `image.playableDuration` (for every video).

Excluding `playableDuration` in particular improves performance significantly when the user has videos.

This change:

- Adds `include` params for `image.width`, `image.height`, and `image.playableDuration`
- Refactors Android code to respect those and avoid looping if the params are excluded
- Makes iOS code respect it too
- Adds documentation

Side change:

- This also deletes the useless `isStored` output that's just a boolean always set to true on iOS

## Test Plan

Added toggles to GetPhotosPerformanceExample.

| iOS | Android |
| --- | --- |
| Camera iOS | Camera Android |

Just the standard `yarn start:ios` and `yarn start:android`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
@@ -517,7 +517,8 @@ public class CameraRollModule extends ReactContextBaseJavaModule {
       WritableMap node = new WritableNativeMap();
       boolean imageInfoSuccess =
           putImageInfo(resolver, media, node, widthIndex, heightIndex, sizeIndex, dataIndex,
-              mimeTypeIndex, includeFilename, includeFileSize, includeImageSize, includePlayableDuration);
+              mimeTypeIndex, includeFilename, includeFileSize, includeImageSize,
+              includePlayableDuration);
       if (imageInfoSuccess) {
         putBasicNodeInfo(media, node, mimeTypeIndex, groupNameIndex, dateTakenIndex);
         putLocationInfo(media, node, dataIndex, includeLocation);
@@ -545,6 +546,10 @@ public class CameraRollModule extends ReactContextBaseJavaModule {
     node.putDouble("timestamp", media.getLong(dateTakenIndex) / 1000d);
   }
 
+  /**
+   * @return Whether we successfully fetched all the information about the image that we were asked
+   * to include
+   */
   private static boolean putImageInfo(
       ContentResolver resolver,
       Cursor media,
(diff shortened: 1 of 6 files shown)


bartolkaruza:
@hsource the other PR is merged, happy to help merging this

hsource:
Thanks! I rebased the change and retested it, and updated the GIFs in the original post! Take a look. It should be ready for review and submission @bartolkaruza

android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
```diff
@@ -583,25 +589,31 @@ private static boolean putImageInfo(
...
       boolean includePlayableDuration) {
+    image.putNull("playableDuration");
+
     if (!includePlayableDuration || !isVideo) {
-      image.putNull("playableDuration");
-      return;
+      return true;
     }
 
+    boolean success = true;
```
bartolkaruza, line 610: Do you have an example of how/when you would use this returned boolean in user code?

hsource: I don't think this will ever be called directly user code, because these functions are only called by `putEdges`. `putEdges` is called as part of `getPhotos`.

The previous code for [`putEdges`](https://github.com/react-native-community/react-native-cameraroll/blob/master/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java#L513) skipped over images where `putImageInfo` returned false. This boolean ensures that the same logic applies.

`putImageInfo` used to return false whenever it failed to fetch the `playableDuration` ([line 595](https://github.com/react-native-community/react-native-cameraroll/blob/master/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java#L595)) or the `width`/`height` ([line 612](https://github.com/react-native-community/react-native-cameraroll/blob/master/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java#L612)).

If we fail to fetch either of them, we assume something's wrong with the file and return false. This would cause `putPlayableDurationSuccess` or `putImageSizeSuccess` to be set to false inside of `putImageInfo`. `putImageInfo` would then return false, which would cause the code in `putEdges` to skip the image, just like before.

bartolkaruza: right, that's not a user concern then, thanks for the answer.

bartolkaruza:
Thank you for contributing @hsource !

Issue #194, opened 2020-06-08
Document Android request permissions


bartolkaruza:
This issue is made to keep track of adding the request for permissions that is needed on Android 29 to the documentation. The workaround mentioned on #19 describes in more detail what is required.

https://github.com/react-native-community/react-native-cameraroll/issues/19

kiranjd:
I would like to take this up

kiranjd:
#201 PR resolves this @bartolkaruza

bartolkaruza:
This is now fixed on master

PR #202, opened 2020-06-19
Adds docs for Android permissions

kiranjd:
# Summary
Adds the missing bit where you have to explicitly ask permission.

#194 
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Permission is required to read and write to the external storage.
 
 On Expo, follow the guide [here](https://docs.expo.io/versions/latest/sdk/permissions/) for requesting the permission.
 
-On ejected apps, adding the following lines will add the capability for the app to request the permission. Find more info on Android Permissions [here](https://reactnative.dev/docs/permissionsandroid).
+On react-native-cli or ejected apps, adding the following lines will add the capability for the app to request the permission. Find more info on Android Permissions [here](https://reactnative.dev/docs/permissionsandroid).
 
 ```xml
 <manifest>
--- a/README.md
+++ b/README.md
@@ -63,9 +63,15 @@ The user's permission is required in order to access the Camera Roll on devices 
 
 If you are targeting devices running iOS 11 or later, you will also need to add the `NSPhotoLibraryAddUsageDescription` key in your `Info.plist`. Use this key to define a string that describes how your app will use this data. By adding this key to your `Info.plist`, you will be able to request write-only access permission from the user. If you try to save to the camera roll without this permission, your app will exit.
 
-**Andriod**
+**Android**
 
-Permission is required to read and write to the external storage. Adding the following lines will add the capability for the app to request for perission.
+This code only works with ejected apps, and not with Expo React Native apps. I suggest also pointing users via a link to both react-native's PermissionsAndroid and Expo's Permissions APIs in addition to including sample code.
+
+Permission is required to read and write to the external storage.
+
+On Expo, follow the guide [here](https://docs.expo.io/versions/latest/sdk/permissions/) for requesting the permission.
+
+On ejected apps, adding the following lines will add the capability for the app to request the permission. Find more info on Android Permissions [here](https://reactnative.dev/docs/permissionsandroid).
 
 ```xml
 <manifest>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,14 +57,17 @@ import CameraRoll from "@react-native-community/cameraroll";
...
 
+**iOS**
+
 The user's permission is required in order to access the Camera Roll on devices running iOS 10 or later. Add the `NSPhotoLibraryUsageDescription` key in your `Info.plist` with a string that describes how your app will use this data. This key will appear as `Privacy - Photo Library Usage Description` in Xcode.
 
 If you are targeting devices running iOS 11 or later, you will also need to add the `NSPhotoLibraryAddUsageDescription` key in your `Info.plist`. Use this key to define a string that describes how your app will use this data. By adding this key to your `Info.plist`, you will be able to request write-only access permission from the user. If you try to save to the camera roll without this permission, your app will exit.
 
-On Android permission is required to read the external storage. Add below line to your manifest to request this permission on app install.
-On Android permission is required to write in the external storage when saving to camera roll. Add below line to your manifest to request this permission on app install.
+**Andriod**
```
hsource, line 66: Minor typo - "Android"

README.md
```diff
@@ -57,14 +57,17 @@ import CameraRoll from "@react-native-community/cameraroll";
...
 
-```
+Permission is required to read and write to the external storage. Adding the following lines will add the capability for the app to request for perission.
```
hsource, line 68: Minor typo and grammar suggestion: "add the capability for the app to request for perission" -> "add the capability for the app to request the permission"

README.md
```diff
@@ -73,6 +76,38 @@ On Android permission is required to write in the external storage when saving t
 <application>
 ```
 
+Then you have to explicitly ask for the permission
```
hsource, line 79: This code only works with ejected apps, and not with Expo React Native apps. I suggest also pointing users via a link to both [react-native's PermissionsAndroid](https://reactnative.dev/docs/permissionsandroid) and [Expo's Permissions](https://docs.expo.io/versions/latest/sdk/permissions/) APIs in addition to including sample code.

README.md
```diff
@@ -73,6 +76,38 @@ On Android permission is required to write in the external storage when saving t
...
+
+```javascript
+import { PermissionsAndroid, Platform } from "react-native";
+import CameraRoll from "@react-native-community/cameraroll";
+
+const checkAndroidPermission = async () => {
```
hsource, line 85: If we still wanted to include some sample code (which I don't think is needed), here's how I'd restructure this code to be more succinct. It solves 2 specific problems:

1. Promise.resolve() does nothing in an async function
2. We still unnecessarily call the `saveToCameraRoll` function even if we don't have the permission

```js
import { PermissionsAndroid, Platform } from "react-native";
import CameraRoll from "@react-native-community/cameraroll";

async function hasAndroidPermission() {
  const permission = PermissionsAndroid.PERMISSIONS.WRITE_EXTERNAL_STORAGE;
  let hasPermission = await PermissionsAndroid.check(permission);
  if (hasPermission) {
    return true;
  }
  const status = await PermissionsAndroid.request(permission);
  return status === 'granted';
}

async function savePicture() {
  if (Platform.OS === "android" && !(await hasAndroidPermission())) {
    return;
  }
  CameraRoll.saveToCameraRoll(tag, [type]);
};
```

kiranjd: Yes, that makes sense. Thanks. I would want to change `let` to `const` though as it serves no purpose being a `let`.

README.md
kiranjd: Thanks for the links.

Do you mean I should include Expo's permission code snippet as well?

kiranjd: I hope the changes reflect what is being expected. Let me know if you want if something is out of place.

hsource: Good catch on the `let`. This looks good!

README.md
```diff
@@ -57,14 +57,23 @@ import CameraRoll from "@react-native-community/cameraroll";
...
+**Android**
 
-```
+This code only works with ejected apps, and not with Expo React Native apps. I suggest also pointing users via a link to both react-native's PermissionsAndroid and Expo's Permissions APIs in addition to including sample code.
+
+Permission is required to read and write to the external storage.
+
+On Expo, follow the guide [here](https://docs.expo.io/versions/latest/sdk/permissions/) for requesting the permission.
+
+On ejected apps, adding the following lines will add the capability for the app to request the permission. Find more info on Android Permissions [here](https://reactnative.dev/docs/permissionsandroid).
```
bartolkaruza, line 74: Minor nitpick: `On non-expo or ejected apps`
This is because many apps don't start with expo, and eject, but start with react-native-cli instead.

kiranjd: Got it. Resolved and pushed.

bartolkaruza:
@kiranjd thank you for contributing!

PR #209, opened 2020-06-30
Remove unnecessary publish job requirement

cesargdm:
# Summary

Cleanup of the CircleCI workflow, there's no need to make publish job dependant of analyze, since analyze is already dependent on checkout_code

Screen Shot 2020-06-30 at 1 21 46 PM

New pipeline:
Screen Shot 2020-06-30 at 1 25 28 PM

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,7 +80,8 @@ workflows:
       #       - analyze
       - publish:
           requires:
-            - rn/android_build
+            - android_debug_build
+            - android_release_build
           filters:
             branches:
               only: master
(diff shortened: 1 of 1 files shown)


cesargdm:
Actually, does publish would need android_* jobs to be completed before?

bartolkaruza:
> there's no need to make publish job dependant of analyze
Did you mean `there is no need to make publish depend on checkout_code`?

bartolkaruza:
> Actually, does `publish` would need `android_*` jobs to be completed before?

It would be an improvement. The current setup is left over from when the android builds were very unreliable. Right now, they are passing consistently already for a long time, so we could make them required to pass before publishing.

PR #215, opened 2020-07-10
Update README.md

greg7gkb:
Clarify import statement for TypeScript projects

# Summary
Clarification only in README for TypeScript builds.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,12 @@ to:
 import CameraRoll from "@react-native-community/cameraroll";
 ```
 
+or in TypeScript:
+
+```typescript
+import * as CameraRoll from '@react-native-community/cameraroll';
+```
+
 ## Usage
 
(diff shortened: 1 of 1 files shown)


hsource:
Is this necessary when you have `esModuleInterop` off? I have `"esModuleInterop": true` in my tsconfig and this package works just fine with that. I believe the default [React Native Typescript config](https://github.com/react-native-community/react-native-template-typescript/blob/master/template/tsconfig.json#L46) also has `esModuleInterop` set to true, so I think it's fine to leave the docs as is.

greg7gkb:
I think it's necessary... our tsconfig does not specify esModuleInterop, so whatever the default is is not working.

Issue #233, opened 2020-09-16
Support React Native v0.63.


chrisbobbe:
Like https://github.com/react-native-community/react-native-cameraroll/issues/127, but for React Native v0.63. RN v0.63 was [released](https://reactnative.dev/blog/2020/07/06/version-0.63) on 2020-07-06. There don't appear to be any [announced breaking changes](https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#v0630) that would affect this library, but I might have missed some, and in any case thorough testing should be done.

- The upper bound for `"react-native"` under `peerDependencies` should be increased to include 0.63.
- `"react-native"` in `devDependencies` should be set to the current latest 0.63, and `"react"` and `"flow-bin"` (maybe more) should be updated to match whatever the `react-native` repo has for those at its current latest 0.63.
- If things break, adjustments to the code might be necessary.
  - If these adjustments break compatibility with older versions of React or React Native that are within their previous range under `peerDependencies`, then the range should be narrowed to exclude those versions, and it becomes a breaking change. (This is easy to miss; see https://github.com/semver/semver/issues/502#issuecomment-466501843, and, e.g., https://github.com/react-native-community/react-native-webview/issues/1445#issue-635011818.)
  - Before making a release, do a sanity check on the peer-dep ranges for `react-native` and `react`, keeping in mind that every React Native version (that I've seen) is pinned to a single React version. E.g., see https://github.com/react-native-community/react-native-webview/issues/1445#issuecomment-642887123, where I pointed out that a React peer-dep range of ^16.9 effectively means that React Native versions before 0.61 are not supported, even if the `react-native` peer-dep range includes versions before 0.61.

fairlycasual:
Thanks @chrisbobbe, this would be very useful for our organization as well as we are in the middle of upgrading RN from 0.61.4 to 0.63.2, and have come to this point blocking us on iOS. We will attempt to fork the library into our own repo and make these changes to see if they solve the issues we are facing, and if we are successful could open a PR for the changes. Otherwise, I think we can achieve the same save functionality using RNFS utilities.

fairlycasual:
I should note that no alterations to the package were necessary in the end during our upgrade process. We just needed to create a bridging header to let Swift and Objective-C libraries communicate together. Now all works as expected. A quick resource can be found at: 
https://stackoverflow.com/a/56187043

fairlycasual:
[...]

PR #260, opened 2020-11-30
fix: updated react dependency in podspec to enable build in Xcode 12 (for iOS >= 12)

alexisbronchart:
See https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "license": "MIT",
   "peerDependencies": {
     "react": "^16.0",
-    "react-native": ">=0.57 <=0.62"
+    "react-native": ">=0.60 <=0.62"
   },
   "devDependencies": {
     "@babel/core": "^7.9.6",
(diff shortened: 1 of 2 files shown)


karvulf (approved):

chrisbobbe:
Just a heads-up 🙂: this would break compatibility with RN v0.59 and earlier because `React-Core` doesn't exist as its own  pod until https://github.com/facebook/react-native/commit/2321b3fd7, which was released in RN v0.60.

The range for `react-native` under `peerDependencies` in package.json currently (as of 4.0.1; [link](https://github.com/react-native-cameraroll/react-native-cameraroll/blob/v4.0.1/package.json#L42)) contains versions that would no longer work; I would recommend narrowing that range so that it excludes them, and announcing this as a breaking change in a major-version release.

package.json
```diff
@@ -39,7 +39,7 @@
   "license": "MIT",
   "peerDependencies": {
     "react": "^16.0",
-    "react-native": ">=0.57 <=0.62"
+    "react-native": ">=0.60 <=0.63"
```
chrisbobbe, line 42: I don't think the `React` to `React-Core` change lets us raise this upper bound to 0.63; I think that would best be done as part of #233 (I've just opened #264 for that).

alexisbronchart: Allright it's back to 0.62

Naturalclar (approved):
@alexisbronchart @chrisbobbe Thanks!

PR #264, opened 2020-12-03
chore: Support React Native v0.63.

chrisbobbe:
# Summary

Fixes: #233

## Test Plan

- [x] Run `yarn test`
- [x] Run example app with `yarn start:android` and `yarn start:ios` and see that nothing is obviously broken
- [ ] Run Detox tests with `yarn test:detox:...`—**I'm not sure how to work with Detox; I haven't done it before.** Running one of those (`test:detox:android:test:debug`) failed for me before my changes, at `master`; is this expected?
- [ ] Point an app in the wild, using RN v0.63, to the updated package to make sure it works—**I'm not sure how best to do this.** Should I expect it to work if I just point `@react-native-community/cameraroll` in my app's `package.json` at the path of my local Git clone (i.e., `"@react-native-community/cameraroll": "../react-native-cameraroll",`)? I tried this, and my app failed just after opening, with a few errors like `ExceptionsManager.js:76 Invariant Violation: Module HMRClient is not a registered callable module (calling setup)`. This was the case even on `master`, before my changes, so I suspect there might be a build step that gets done when publishing to NPM, but I can't find it.
- [x] Point an app in the wild, using RN v0.60 (and some other versions in the [">=0.60 <=0.62" range](https://github.com/react-native-cameraroll/react-native-cameraroll/blob/4f16a5a/package.json#L42) if we want to be extra thorough), to the updated package to make sure we haven't accidentally dropped support for RN versions we claim to support—**(see previous item)**

(See "Test Plan", above.)

(See "Test Plan", above.)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] I have tested this on a simulator (iOS)
- [ ] I have tested this on a simulator (Android)
- [ ] I have tested this on an iOS device—Xcode was asking for a key (don't remember which) before I could run on a device; I'm happy to go through that if you'd like.
- [x] I have tested this on an Android device
- [ ] ~~I added the documentation in `README.md`~~ (compatible RN versions aren't already listed in the README)
- [ ] ~~I updated the typed files (TS and Flow)~~ (nothing to update)
- [x] I added a sample use of the API in the example project (`example/App.js`) (I updated the sample app to work with RN v0.63.)


MichaelDanielTom:
What do you all think about removing the upper bound peer dependency of `<=0.63` entirely, just to make forward compatibility easier, so we don't need to rely on making changes with each new version of react-native?

There is the downside of not being sure that it would work with a new version, IMO the pros outweigh the cons.

Hugogsalas:
So, there is no support yet?

chrisbobbe:
> What do you all think about removing the upper bound peer dependency of `<=0.63` entirely, just to make forward compatibility easier, so we don't need to rely on making changes with each new version of react-native?
> 
> There is the downside of not being sure that it would work with a new version, IMO the pros outweigh the cons.

Hi, sorry for the delay in responding! I would favor keeping the upper bound; here's why.

- In my experience, it's not unusual for things in a package to break with a new React Native version. Without the upper bound, we'd be saying (as part of the module's public API) that it works with all future React Native versions—which, I suppose, includes changes that no one right now knows anything about. It's a claim that invites skepticism. (If we removed the upper bound, I'd say it would be smart for the README to say, prominently, something like "We expect it to work with future React Native versions, _but please report any issues with RN versions we might not have tested yet_." A statement like that would boost confidence that we know things might break and that we'll try to fix them if they do.)
  - There might be breakages specific to the kind of work this package does.
  - There might be breakages that could affect just about anything in the RN ecosystem. For example, https://github.com/facebook/react-native/commit/2321b3fd7, which was released in RN v0.60, broke libraries that kept a line like `s.dependency 'React/Core'` in their Podspec, instead of `s.dependency 'React'` or `s.dependency 'React-Core'`. For projects that hadn't seen that coming (and part of the point is that it was hard to see coming), it meant some pretty urgent debugging—e.g., https://github.com/zmxv/react-native-sound/pull/579, replaced by https://github.com/zmxv/react-native-sound/pull/585, to resolve https://github.com/zmxv/react-native-sound/issues/584; https://github.com/joltup/rn-fetch-blob/pull/397, to resolve https://github.com/joltup/rn-fetch-blob/issues/365.
  - As soon as a new React Native version that breaks something in `react-native-cameraroll` is released, all N releases of `react-native-cameraroll` that omit the RN upper bound would be omitting it in error (i.e., saying the new version is supported when it actually isn't). That means a lot of (preventable) errors, if N is large—and N would naturally grow over time.
- We already rely on making changes with each new version of React Native (or, I think that's a goal, anyway). In particular, there's an effort to keep the example app up-to-date with the latest version of React Native (e.g., this PR updates the example app). As long as we're keeping up with that, I think it wouldn't be too much extra work to run some tests and bump the upper bound for `react-native`.

Naturalclar (approved):
@chrisbobbe thanks!
confirmed that example works for both iOS and Android

PR #275, opened 2020-12-30
XCode 12 compatibility fix

slobodans79:
# Summary

Latest Xcode 12 fails to build while without a module to depend on React-Core directly hence this change is necessary for all native modules on iOS. For more details please check: https://github.com/facebook/react-native/issues/29633

## Test Plan

Use this branch to install with an app running on Xcode 12.


chrisbobbe:
I believe this is a duplicate of #260. One potentially useful note that exists there, but not here, is https://github.com/react-native-cameraroll/react-native-cameraroll/pull/260#issuecomment-737494056.

Naturalclar:
@slobodans79 thanks for the PR! the change has already landed on master with #260, so I'll be closing this PR

PR #277, opened 2021-01-04
fix: (Android) Fixed issue where timestamp was "0"

hannojg:
# Summary

On android, we discovered an issue where the "timestamp" would be just `"0"` on several test devices. 
To solve this we added this:
- When dateTaken is 0 retrieve value from dateAdded
- Added additional modified field.

## Test Plan

We tested this on several physical test devices as well as on emulators.

Can't exactly tell how to reproduce the issue, that the dateTaken was 0 on some devices.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist


billnbell:
this seems like a no brainer lets approve.

Naturalclar (approved):
@hannojg thanks!

PR #278, opened 2021-01-07
Added orientation (Android only)#212

### LINKED ISSUE #212: When photo/video has rotation the imageSize property will be wrong on Android
Hi, I'm using the latest version 4.0.0 and when I use `CameraRoll.getPhotos` to fetch images and videos with rotation (90 degrees) the width and height will be opposite.

Here is the info when using ffprob on a video file:

```
    Stream #0:0(und): Video: hevc (Main) (hvc1 / 0x31637668), yuv420p(tv, bt709), 1920x1080, 11409 kb/s, 60 fps, 60 tbr, 600 tbn, 600 tbc (default)
    Metadata:
      rotate          : 90
      creation_time   : 2020-07-05T15:11:27.000000Z
      handler_name    : Core Media Video
      encoder         : HEVC
```

I'm getting in the image properties width 1920 and height 1080 which ok I guess, but there is no way for me to know the video is rotated by 90 degrees.

Any workaround for it?

MarcoScabbiolo:
# Summary

Added optional `orientation` property to image included by `orientation`. Works for Android only and provides the necessary information to resolve #212 in user-land

**Not implemented in iOS**, the underlying issue addressed by this PR doesn't happen in iOS and I personally develop in Swift for iOS, not Objective-C.

## Test Plan

Tested it on a real device, already have a patch-packge in place in my app and this will go through our QA process and end up in production independently of this PR. Would gladly remove patch and update when this PR is merged though!

Having a rotated video in the device, can be done by just recoding a landscape video with the device's camera, this should produce a 90/270 orientation

Having the previously mentioned video, it will be retrieved along with its corresponding `orientation` value, if `orientation` is included. In user-land it is posible to detect a 90/270 rotation and swap the already swapped dimensions to have the real ones.

For example, a 400x500 video recorded in landscape would have
```javascript
{ width: 500, height: 400, orientation: 90 }
```

Video players like `react-native-video` understand the rotation and reproduce it properly but without `orientation` it is imposible to know before-hand if the video is actually rotated and things like advanced layouts and file uploads will have the dimensions swapped.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅    |

## Checklist

- [ X ] I have tested this on a device and a simulator
- [ X ] I added the documentation in `README.md`
- [ X ] I updated the typed files (TS and Flow)

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,9 @@ Returns a Promise with photo identifier objects from the local camera roll of th
   * `location`: Ensures `location` is available in each node. This has a large performance impact on Android.
   * `imageSize` : Ensures `image.width` and `image.height` are available in each node. This has a small performance impact on Android.
   * `playableDuration` : Ensures `image.playableDuration` is available in each node. This has a medium peformance impact on Android.
+  * `orientation`: Ensures `image.orientation` is available in eacho node.
+  This has a low performance impact on Android.
+  Works for Android only
 
 Returns a Promise which when resolved will be of the following shape:
 
@@ -222,6 +225,7 @@ Returns a Promise which when resolved will be of the following shape:
       * `width`: {number | null} : Only set if the `include` parameter contains `imageSize`
       * `fileSize`: {number | null} : Only set if the `include` parameter contains `fileSize`
       * `playableDuration`: {number | null} : Only set for videos if the `include` parameter contains `playableDuration`. Will be null for images.
+      * `orientation`: {number | null} : Only set if the `include` parameter contains `orientation`. *Android only*
     * `timestamp`: {number}
     * `location`: {object | null} : Only set if the `include` parameter contains `location`. An object with the following shape:
       * `latitude`: {number}
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
@@ -48,6 +48,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.channels.FileChannel;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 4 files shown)


Arjan-Zuidema:
@MarcoScabbiolo Please resolve the conflicts so this can be tested

zedtux:
@MarcoScabbiolo please, this PR is really needed. Thank you 🙏

zedtux:
To let you all know, I have applied manually this patch on my local copy of this library from my `node_modules/` folder, and I found one missing thing to get this patch working fine.

The following line should be added in the file `src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java` to the `PROJECTION` string array :

```diff
@@ -94,7 +96,8 @@ public class CameraRollModule extends ReactContextBaseJavaModule {
           MediaStore.MediaColumns.WIDTH,
           MediaStore.MediaColumns.HEIGHT,
           MediaStore.MediaColumns.SIZE,
-          MediaStore.MediaColumns.DATA
+          MediaStore.MediaColumns.DATA,
+          MediaStore.MediaColumns.ORIENTATION
   };
 
   private static final String SELECTION_BUCKET = Images.Media.BUCKET_DISPLAY_NAME + " = ?";
```

Then I was able to retrieve the photos orientation and apply a rotation making them appearing correctly in my app.

Arjan-Zuidema:
@zedtux Can you create a PR from this PR? As you already have a working patch. I'll be happy to merge your PR as @MarcoScabbiolo seems to be inactive

zedtux:
Well, in the case you promise to merge, then I can do it next week, for course !

HyopeR:
Photos fetched on Android are not always s**orted by width-height**. So it makes sense to find an orientation in the background.

But instead of returning the orientation to the user, I think we should **rearrange the correct width and height** values and return this to the user.

Thus, **the keys and values will always remain true** and the user can simply do the following, for example; width > height = landscape.

Mats01-fer:
I've encountered the same problem on Android and came across this. I have since discovered that, while this doesn't happen on iOS devices, it does happen in the **iOS simulator**. 

Is sorting **width-height** always the solution?

In any case, returning the **orientation** to the user might still be useful for other use-cases since it's available.

PR #281, opened 2021-01-13
allow react-native 0.63.x

sapieneptus:
# Summary

Following up on #233 - extends peer dependency range to include 0.63. 

## Test Plan

`npm run test` succeeds. Incorporated this change (from my fork) into an app locally and was able to save photos to the device & simulator. Please advise if there's something more I should do to demonstrate that it works. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅   |

## Checklist

~- [ ] I added the documentation in `README.md`~ N/A
~- [ ] I updated the typed files (TS and Flow)~ N/A
~- [ ] I added a sample use of the API in the example project (`example/App.js`)~ N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "license": "MIT",
   "peerDependencies": {
     "react": "^16.0",
-    "react-native": ">=0.57 <=0.62"
+    "react-native": ">=0.57 <=0.63"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


chrisbobbe:
This is a duplicate of #264.

PR #303, opened 2021-04-12
Append file extension in the type field for iOS

Northen-Light:
# Summary

Issue : https://github.com/react-native-cameraroll/react-native-cameraroll/issues/301#issue-854483787
Idea taken : https://github.com/react-native-cameraroll/react-native-cameraroll/issues/12#issue-422708927

In the current version, react-native-cameraroll does not provide file extension for the photos and images taken from "photo" app of iOS devices. To fix that I have appended the file extension in the type key.

Current implementation :  
1. for photos, type key is like this, "type" : image
2. for videos, type key is like this, "type" : video

For more details, have a look into the issue https://github.com/react-native-cameraroll/react-native-cameraroll/issues/301#issue-854483787

After this change: 
1. for photos, type key would be like this, "type" : image/png  | image/jpeg | image/heic | ...
2. for videos, type key would be like this, "type" : video/mov | video/mpeg-4 | ...

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

I have the tested the changes in the simulator and iOS devices as well.

### DIFF
--- a/ios/RNCCameraRollManager.m
+++ b/ios/RNCCameraRollManager.m
@@ -359,13 +359,23 @@ RCT_EXPORT_METHOD(getPhotos:(NSDictionary *)params
         return;
       }
 
-      NSString *const assetMediaTypeLabel = (asset.mediaType == PHAssetMediaTypeVideo
+      NSString *assetMediaTypeLabel = (asset.mediaType == PHAssetMediaTypeVideo
                                             ? @"video"
-                                            : (asset.mediaType == PHAssetMediaTypeImage
+                                             : (asset.mediaType == PHAssetMediaTypeImage
                                                 ? @"image"
                                                 : (asset.mediaType == PHAssetMediaTypeAudio
                                                   ? @"audio"
                                                   : @"unknown")));
+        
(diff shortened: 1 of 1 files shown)


pribeh:
Thanks @Northen-Light. We require this, can someone please review and pass it. We've tested it and works as designed

mishsom:
@bartolkaruza can we merge this if you don't see an issue?

mishsom (approved):

ios/RNCCameraRollManager.m
```diff
@@ -374,6 +384,7 @@ static void RCTResolvePromise(RCTPromiseResolveBlock resolve,
           @"group_name": currentCollectionName,
           @"image": @{
               @"uri": uri,
+              @"extension": extension,
```
mishsom: removed in next commit

mishsom (approved):

Northen-Light:
This implemenation does not correctly provide all possible type of video extensions.

PR #307, opened 2021-04-27
new MediaStore api for Android 10 and above

HugoGresse:
# Summary

This should fix the issue with ScopedStorage on android 10 and above by using the MediaStore API. 
It's only made for the `save` part of CameraRoll and has not yet been tested in production. 

Missing part: unique file name

## Test Plan

No test plan yet, feel free to test the PR so we can improve it together. 

Android 10+

Take a new picture (from Camera for eg), and use the save on CameraRoll. The file/image should be visible in any device Gallery apps as well as the correct folder from DCIM or Pictures on the device itself. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/Utils.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/Utils.java
@@ -1,0 +1,38 @@
+package com.reactnativecommunity.cameraroll;
+
+import android.content.ContentResolver;
+import android.content.Context;
+import android.database.Cursor;
+import android.net.Uri;
+import android.provider.MediaStore;
+
+import javax.annotation.Nullable;
+
+public class Utils {
+
+  // From https://stackoverflow.com/a/64359655/1377145
+  @Nullable
+  public static String getFilePathFromContentUri(Context context, Uri contentUri) {
+    ContentResolver contentResolver = context.getContentResolver();
+    Cursor cursor = contentResolver.query(contentUri, null, null, null, null);
+    if (cursor == null) {
+      return null;
+    }
+    cursor.moveToFirst();
+    String document_id = cursor.getString(0);
+    document_id = document_id.substring(document_id.lastIndexOf(":") + 1);
+    cursor.close();
+
+    cursor = contentResolver.query(
+            android.provider.MediaStore.Images.Media.EXTERNAL_CONTENT_URI,
+            null, MediaStore.Images.Media._ID + " = ? ", new String[]{document_id}, null);
+    if (cursor == null) {
+      return null;
+    }
+    cursor.moveToFirst();
+    String path = cursor.getString(cursor.getColumnIndex(MediaStore.Images.Media.DATA));
+    cursor.close();
+    return path;
+  }
+
+}
--- a/README.md
+++ b/README.md
@@ -37,15 +37,8 @@
   	```
       compile project(':@react-native-community_cameraroll')
   	```
-Starting with Android 10, the concept of [scoped storage](https://developer.android.com/training/data-storage#scoped-storage) is introduced. Currently, to make it working with that change, you have to add `android:requestLegacyExternalStorage="true"` to `AndroidManifest.xml`:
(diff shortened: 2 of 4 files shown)


dokkis:
@HugoGresse 

have you checked https://developer.android.com/reference/android/content/Context#getExternalFilesDir(java.lang.String) ?

seems like there are also some example with the MediaScannerConnection, seems closer to what the library does today and would be simpler if works (?), not sure :)

HugoGresse:
> @HugoGresse
> 
> have you checked https://developer.android.com/reference/android/content/Context#getExternalFilesDir(java.lang.String) ?
> 
> seems like there are also some example with the MediaScannerConnection, seems closer to what the library does today and would be simpler if works (?), not sure :)

I'm not sure to understand correctly, I've reused the MediaScanner from this repo.

dokkis:
@HugoGresse never mind, I thought/hope we could just use getExternalFilesDir as opposed to getExternalStoragePublicDirectory and keep all the code as is, but I tried and seems like the MediaScanner is not really working in that case from some tests I have made, anyway also take a look at an answer I gave here: https://github.com/status-im/status-react/issues/12063. I have some suspect we don't really need to change anything and the campaign is run a bit poorly with lack of information.

marcshilling:
I'm getting a crash on this line in `getFilePathFromContentUri`:
`document_id = document_id.substring(document_id.lastIndexOf(":") + 1);`
>Attempt to invoke virtual method 'int java.lang.String.lastIndexOf(java.lang.String)' on a null object reference

Because `document_id` is null. I'm attempting to save an image taken via camera from the emulator. The contentUri being passed in is `content://media/external/images/media/39`

demedos:
Any update on this? @HugoGresse 
I have the same issue as @marcshilling . 
Also, it's a bit confusing about whether one should use the code provided in this MR, the one in the comment https://github.com/react-native-cameraroll/react-native-cameraroll/issues/304#issuecomment-842383234 or the one on [SO](https://stackoverflow.com/questions/67179409/alternative-for-mediascanner-scanfile-cameraroll-on-android-29-q/67179410#67179410)

HugoGresse:
Sorry for the lack of update. I've been using my own module while CameraRoll is updated. 
I've compared both the 3 links you posted and the production code, none check for a null `document_id`.  However, where those are different is that I'm not using the code around the specific `moveToMediaStore` used in this repo (which is most of the refacto of this PR), meaning I've never tested this PR in production... 
I don't have time to work on updating the PR for now, but if you want to take it from here and open your own PR for this one, feel free so. 
Do you have example where document_id is null? Maybe it's related to the initial url/uri provided?

demedos:
Replacing the code in `getFilePathFromContentUri` from [the current solution](https://stackoverflow.com/a/64359655) to [this](https://stackoverflow.com/a/11603899) fixed the document_id issue. I don't know if it's worth making an example app to investigate the reason it isn't working right now.

android/src/main/java/com/reactnativecommunity/cameraroll/Utils.java
```diff
@@ -0,0 +1,38 @@
...
+import android.net.Uri;
+import android.provider.MediaStore;
+
+import javax.annotation.Nullable;
+
+public class Utils {
+
+  // From https://stackoverflow.com/a/64359655/1377145
+  @Nullable
+  public static String getFilePathFromContentUri(Context context, Uri contentUri) {
```
demedos, line 15: Replacing this with [this other answer](https://stackoverflow.com/a/11603899) on the same SO post fixes the null document_id issue

HugoGresse: can you make sure it works on android9, 10 and 11?

demedos: I'll see if I have time to do so. right now I'm using a custom native module just like yours, which checks the api target level and defaults to CameraRoll if it is < 11

idrissakhi:
Resolved in v5.0.0

PR #343, opened 2021-09-10
Fix image dimensions info when the file has some EXIF orientation (Android)

drash-course:
# Summary

In our App, we want to display only the landscape (horizontal) pictures from the user's gallery. To do so, we check the `width` and `height` data given by the lib when calling `CameraRoll.getPhotos(...)`. Unfortunately the data is incorrect on some devices (e.g. Xiaomi Redmi Note 5) where width and height are swapped around.

## Root problem

Pictures taken with the native camera app may set the EXIF orientation property instead of rotating the pixel data. When reading the Bitmap data, you need to check the EXIF attributes and apply the required rotation (if any). Currently the lib assumes that there isn't any rotation to be done.

## Solution

I added code in `putImageSize(...)` to check the EXIF data and swap the width / height data if necessary. It will apply the check whether the size data is retrieved from the cursor's columns or by reading the Bitmap data. The performance impact should be very small as we are only loading the file headers and reusing the file handle.

## Test Plan

There is no API change. Testing entails calling `CameraRoll.getPhotos({ include: ['imageSize'], assetType: 'Photos' })` on a device with pictures that have EXIF rotation data. 

You need either:
- A device like the Xiaomi Redmi Note 5 that sets EXIF orientation data when taking pictures 
- A set of pictures with different EXIF orientation data

- Open the example/ project
- Go to example/js/CameraRollView.js
- add `console.log(asset)` at line 180 in `_appendAssets(data)`
- Connect an android device or simulator
- Install and launch the app
- Go to the "CameraRollExample"
- check the logs and make sure the `node.image.width` and `node.image.height` are correct

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    not affected     |
| Android |    ✅     |

## Checklist

- [x] I added the documentation in `README.md` (no change needed)
- [x] I updated the typed files (TS and Flow) (no change needed)
- [x] I added a sample use of the API in the example project (`example/App.js`) (no change needed)

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
@@ -17,6 +17,7 @@ import android.media.MediaMetadataRetriever;
 import android.media.MediaScannerConnection;
 import android.net.Uri;
 import android.os.AsyncTask;
+import android.os.Build;
 import android.os.Environment;
 import android.provider.MediaStore;
 import android.provider.MediaStore.Images;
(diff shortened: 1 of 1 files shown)


hamzaouhssain1993:
such a good developer, such clean code, such a beauty

theasc:
Exactly what I needed, you saved my day burro !!

doranteseduardo (approved):
Changes look OK, tested in a Xiaomi Redmi Note 9 Pro device.

Issue #292, opened 2021-03-03
CameraRoll.save doesn't save on ios 14


JCalds:
I'm facing a problem when using ios 14. 

When I'm trying to save an image using CameraRoll.save() , I'm getting this system popup (which I think isn't the correct one), and consequently, the image is not saved.

![image](https://user-images.githubusercontent.com/79947676/109732854-d5ba8580-7bb5-11eb-85ec-34e78e16a460.png)

In the other iOS versions, the saving process works fine, and I'm getting the correct writeable system popup

Both of the request permissions are needed in diferent contexts NSPhotoLibraryAddUsageDescription, NSPhotoLibraryUsageDescription

Any clue?
Thanks.

JanKaifer:
We don't use `read` access we use only `save`. So we fixed the library here: https://github.com/JanKaifer/react-native-cameraroll

It has probably broken permissions for reading. If someone is interested it should be quite easy to port fix it for reading and writing.

mhammerc:
I have the same problem. On the latest iOS I can't save any image to the cameraroll.

To fix this, with inspiration from @JanKaifer answer and his commit, here is a patch:

```diff
diff --git a/node_modules/@react-native-community/cameraroll/ios/RNCCameraRollManager.m b/node_modules/@react-native-community/cameraroll/ios/RNCCameraRollManager.m
index 4bd76c0..e08e4a1 100644
--- a/node_modules/@react-native-community/cameraroll/ios/RNCCameraRollManager.m
+++ b/node_modules/@react-native-community/cameraroll/ios/RNCCameraRollManager.m
@@ -101,7 +101,7 @@ @implementation RNCCameraRollManager
 static void requestPhotoLibraryAccess(RCTPromiseRejectBlock reject, PhotosAuthorizedBlock authorizedBlock) {
   PHAuthorizationStatus authStatus;
   if (@available(iOS 14, *)) {
-    authStatus = [PHPhotoLibrary authorizationStatusForAccessLevel:PHAccessLevelReadWrite];
+    authStatus = [PHPhotoLibrary authorizationStatusForAccessLevel:PHAccessLevelAddOnly];
   } else {
     authStatus = [PHPhotoLibrary authorizationStatus];
   }
@@ -116,7 +116,7 @@ static void requestPhotoLibraryAccess(RCTPromiseRejectBlock reject, PhotosAuthor
     authorizedBlock(true);
[...]

PR #355, opened 2021-10-22
iOS: Use PHAccessLevelAddOnly for saveToCameraRoll

Arjan-Zuidema:
# Summary

This patch fixes an issue (#292) when trying to save to camera roll on iOS 14 or later. The `requestPhotoLibraryAccess` function requests read and write access on the Photo library. Which triggers a system popup to request access as seen in #292.

The fix adds an extra parameter to determine if `PHAccessLevelReadWrite` or `PHAccessLevelAddOnly` should be used when requesting access to the Photo library.

## Test Plan

This was tested using the example app and our internal app which uses only uses the `saveToCameraroll` feature.

A device or simulator with iOS 14+ and a device or simulator with < iOS 14.

Same as above.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RNCCameraRollManager.m
+++ b/ios/RNCCameraRollManager.m
@@ -98,10 +98,14 @@ static NSString *const kErrorAuthDenied = @"E_PHOTO_LIBRARY_AUTH_DENIED";
 
 typedef void (^PhotosAuthorizedBlock)(bool isLimited);
 
-static void requestPhotoLibraryAccess(RCTPromiseRejectBlock reject, PhotosAuthorizedBlock authorizedBlock) {
+static void requestPhotoLibraryAccess(RCTPromiseRejectBlock reject, PhotosAuthorizedBlock authorizedBlock, bool requestAddOnly) {
   PHAuthorizationStatus authStatus;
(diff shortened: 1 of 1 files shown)


dasaco:
Really looking forward for this to be merged

Dovydastc:
Bump

majirosstefan:
is this lib dead ?

hb-webdev:
Ok if there's no conflicts why hasn't this merged yet? I've been waiting for this implementation

Arjan-Zuidema:
@bartolkaruza anyone active here?

idrissakhi:
@Arjan-Zuidema would you update the ReadMe file please, so users can understand why all permissions are asked when album name is given on save method options.

once updated we can merge

Arjan-Zuidema:
@idrissakhi There already is a sections about iOS [Permissons](https://github.com/react-native-cameraroll/react-native-cameraroll/blob/master/README.md#permissions) in the readme. I think this is clear enough already. (I guess you agree since you merged it to master 😛)

Issue #352, opened 2021-10-20
undefined Unable to resolve module fbjs/lib/invariant


septe01:
when I'm use react-native-cameraroll

getting error :

ERROR  [Error: undefined Unable to resolve module fbjs/lib/invariant from /Users/septehabudin/Desktop/PROJECT/MOBILE/RN/pendekin/node_modules/@react-native-community/cameraroll/js/CameraRoll.js: fbjs/lib/invariant could not be found within the project or in these directories:
  node_modules

If you are sure the module exists, try these steps:
 1. Clear watchman watches: watchman watch-del-all
 2. Delete node_modules and run yarn install
 3. Reset Metro's cache: yarn start --reset-cache
 4. Remove the cache: rm -rf /tmp/metro-*
  12 | import RNCCameraRoll from './nativeInterface';
  13 |
> 14 | const invariant = require('fbjs/lib/invariant');

what do I have to do?
pleas help...

this my package.json 
  "dependencies": {
    "@hookform/resolvers": "^2.8.2",
    "@react-native-async-storage/async-storage": "^1.15.9",
    "@react-native-community/cameraroll": "^4.1.2",
    "@react-native-community/clipboard": "^1.5.1",
    "react": "17.0.2",
    "react-hook-form": "^7.17.5",
    "react-native": "0.66.1"
  },

vfasky:
yarn add fbjs

idrisssakhi:
please check this lib, it provides all same things => react-native-photo-gallery-api https://www.npmjs.com/package/react-native-photo-gallery-api

santhoshvai:
this is fixed as of 5.0.0

talaikis:
> this is fixed as of 5.0.0

Would be nice to have `5.0.0` on npm.

Issue #323, opened 2021-07-07
App failed to run after installing the package (react-native-cameraroll)


KaungHtetHein116:
Error: undefined Unable to resolve module fbjs/lib/invariant from /Users/User/Projects/LibraryTest/node_modules/@react-native-community/cameraroll/js/CameraRoll.js: fbjs/lib/invariant could not be found within the project or in these directories:
  node_modules

If you are sure the module exists, try these steps:
 1. Clear watchman watches: watchman watch-del-all
 2. Delete node_modules and run yarn install
 3. Reset Metro's cache: yarn start --reset-cache
 4. Remove the cache: rm -rf /tmp/metro-*

Issue #317, opened 2021-06-10
Unable to resolve module fbjs/lib/invariant


mkkurt:
Error: Unable to resolve module fbjs/lib/invariant from C:\Users\kutay\Desktop\igRepostRNCli\node_modules\@react-native-community\cameraroll\js\CameraRoll.js: fbjs/lib/invariant could not be found within the 
project or in these directories:
  node_modules
  ..\..\node_modules

I installed the library with npm install and linked with `npx react-native link`
And i'm importing the library like `import CameraRoll from '@react-native-community/cameraroll';`

markw13a:
You could try installing fbjs as per this issue: https://github.com/react-native-cameraroll/react-native-cameraroll/issues/300#issue-852329205

PR #365, opened 2021-12-19
chore: add fbjs as dependency

### LINKED ISSUE #300: Couldn't resolve fbjs
Getting the below error:
> node_modules/@react-native-community/cameraroll/js/CameraRoll.js: fbjs/lib/invariant could not be found within the project or in these directories

On version: 4.0.4
RN 0.64.0

Installing fbjs separately resolves the issue, but this should be documented or included within the package.

UNIDY2002:
# Summary

Fix #300, #317, #323, #352.

As mentioned [here](https://github.com/react-native-cameraroll/react-native-cameraroll/issues/300#issuecomment-940798898), since `fbjs` is indeed a dependency of this library, adding it as a dependency might be a good practice.

This modification shall not impact any part of the library, since the version I tagged for `fbjs` (which is `^1.0.0`) is consistent with that in `yarn.lock`.

## Test Plan

I made a simple test in my own project. Before the modification, my test script will fail, raising an exception as follows:

```
Cannot find module 'fbjs/lib/invariant' from 'node_modules/@react-native-community/cameraroll/js/CameraRoll.js'
```

After the modification, my test script passes.

### Using the current latest version (`4.1.2`)

Screenshot_20211219_154426

### Using the PR version

Screenshot_20211219_154239

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,9 @@
     "photo gallery"
   ],
   "license": "MIT",
+  "dependencies": {
+    "fbjs": "^1.0.0"
+  },
   "peerDependencies": {
     "react": "16 || 17",
     "react-native": ">=0.60"


nvojnovic:
This should be rejected in favour of https://github.com/react-native-cameraroll/react-native-cameraroll/pull/367

idrissakhi:
This is a duplicate with a merged PR 'invariant is being used'

Issue #402, opened 2022-08-11
Retriever.release error


vladimirevstratov:
When i try to run my app on android, this problem appeared on compilation stage.

```
Error: Command failed: ./gradlew app:installDebug -PreactNativeDevServerPort=8081
~/node_modules/@react-native-community/cameraroll/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java:650: error: unreported exception IOException; must be caught or declared to be thrown
      retriever.release();
                       ^
~/node_modules/@react-native-community/cameraroll/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java:713: error: unreported exception IOException; must be caught or declared to be thrown
          retriever.release();
                           ^
Note: ~/node_modules/@react-native-community/cameraroll/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
2 errors

```

React Native 0.69.4
com.android.tools.build:gradle:7.1.1
Repository: mavenCentral instead of jcenter

vladimirevstratov:
Solution:
Replace all `retriever.release();` in `@react-native-community/cameraroll/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java`  with
```
try {
  retriever.release();
} catch (IOException ex) {
  // handle the exception here
}
```

Arjan-Zuidema:
If you could create a pull request with a fix that would be awesome!

vladimirevstratov:
> ~~If you could create a pull request with a fix that would be awesome!~~
> 
> 
> 
> Pull request here #413
> 
> 
> 
> 

Oh, didn't see notification of your response 🙂

santhoshvai:
Note: Make sure the compile SDK version is set to 33. Otherwise, the suggested solution will cause a compilation error.

Pranavpatel428:
> Note: Make sure the compile SDK version is set to 33. Otherwise, the suggested solution will cause a compilation error.

Hello, I have updated my compile SDK version to 33 but still I am getting this error. Also, I am not able to create a build.

PR #413, opened 2022-08-22
fix uncaught exception of retriever.release()

larrylin28:
# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

- [ ] I have tested this on a device and a simulator
- [ ] I added the documentation in `README.md`
- [ ] I updated the typed files (TS and Flow)
- [ ] I added a sample use of the API in the example project (`example/App.js`)

PR #419, opened 2022-08-27
Fix compile error when compileSdkVersion is set to 33

Yonom:
# Summary

Compiling react-native-cameraroll in a react native project with compileSdkVersion set to 33 (Android 13 / Tiramisu) causes a compile error, because IOExceptions from `retriever.release();` are not caught.

This is because starting at API Level 33 `android.media.MediaMetadataRetriever.release()` can throw `IOException`, as noted in the [API Diff Specification](https://developer.android.com/sdk/api_diff/33/changes): 
image

## Test Plan

Compiled our app with the diff applied both with compileSdkVersion 31 and compileSdkVersion 33 to ensure compatibility.

Compile react-native-cameraroll with compileSdkVersion set to 33 in `build.gradle`.

Ensure that the compilation succeeds.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅ (no change)     |
| Android |    ✅      |

## Checklist

- [x] ~I added the documentation in `README.md`~
- [x] ~~I updated the typed files (TS and Flow)~~
- [x] ~~I added a sample use of the API in the example project (`example/App.js`)~~

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
@@ -698,7 +698,11 @@ public class CameraRollModule extends ReactContextBaseJavaModule {
                         + photoUri.toString(),
                 e);
       }
-      retriever.release();
+      try {
+        retriever.release();
+      } catch (IOException e) {
+        // Do nothing. We can't handle this, and this is usually a system problem
+      }
     }
 
     if (photoDescriptor != null) {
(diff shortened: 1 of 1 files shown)


Yonom:
Hmm, I am a bit lost on how to ensure that the library remains compatible with `compileSdkVersion 31` to make the tests pass  - I do not regularly develop in java.

Tests are currently failing in CI because `exception IOException is never thrown in body of corresponding try statement` (which was true until compileSdkVersion 33).

I would appreciate some help here 🙂

santhoshvai:
Hi @Yonom , thanks for this. Just update the compile SDK version to 33 in the library and then the test should pass. 

The path to the file in which it should be changed is `android/gradle.properties`

Yonom:
Thanks @santhoshvai!

santhoshvai:
fixes #402

Arjan-Zuidema:
This is also the same fixes as #413? So, can we close that pull request?

PR #460, opened 2023-02-02
feat: add new arch

WoLewicki:
PR adding Fabric to the library. On `iOS`, there was a need to abstract code from `RNCCameraRollPermission` to `RNCPermissionHelper` since the first one needs to be written in `objc++` which does not handle some of the code which was present there.


android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollPackage.java
```diff
@@ -3,15 +3,104 @@
...
+                                        reactModule.isCxxModule(),
+                                        TurboModule.class.isAssignableFrom(moduleClass)));
+                    }
+
+                    return reactModuleInfoMap;
+                }
+            };
+        } catch (InstantiationException | IllegalAccessException e) {
+            throw new RuntimeException(
+                    "No ReactModuleInfoProvider for com.reactnativecommunity.clipboard.ClipboardPackage$$ReactModuleInfoProvider", e);
```
WoLewicki, line 100:
```suggestion
                    "No ReactModuleInfoProvider for com.reactnativecommunity.cameraroll.CameraRollPackage$$ReactModuleInfoProvider", e);
```

Arjan-Zuidema:
Hey @WoLewicki is this PR done now? 🙂

WoLewicki:
@Arjan-Zuidema there is still some discussion around it here: https://github.com/facebook/react-native/pull/36071, but I think it should be usable now and we could add more changes in another PR. It would be nice for sure if you tested it yourself, but it seemed to work in my use-cases 😅

Arjan-Zuidema:
When loading the FabricExample app I get this error. Looks like the `Image` component is not working properly. I think it's unrelated to this package.

When I dismiss the error and try the other example it seems to work just fine.

Your thought @WoLewicki ?

Simulator Screen Shot - iPhone 14 - 2023-03-03 at 15 51 21

WoLewicki:
@Arjan-Zuidema  yeah I think it might be a problem with the current implementation on new arch, seems like kind of the same issue as discussed in the linked PR.

Arjan-Zuidema:
Ok, will test it again soon. But it seems fine until now 👌🏼

PR #467, opened 2023-02-24
Converting project to kotlin

magom001:
# Summary

Please consider converting the project to kotlin.

Conversion in this PR has been done using Android Studio automatic Java to Kotlin conversion tool.

NO MANUAL CODE CHANGES

Added necessary gradle changes to the package project.

Fixed gradle for the example project.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅    |

## Checklist


android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.kt
```diff
@@ -0,0 +1,1017 @@
...
+                    mediaDetails.put(MediaStore.Images.Media.IS_PENDING, 1)
+                    val resolver = mContext.contentResolver
+                    val mediaContentUri = if (isVideo) resolver.insert(
+                        MediaStore.Video.Media.EXTERNAL_CONTENT_URI,
+                        mediaDetails
+                    ) else resolver.insert(
+                        MediaStore.Images.Media.EXTERNAL_CONTENT_URI,
+                        mediaDetails
+                    )
+                    output = resolver.openOutputStream(mediaContentUri!!)
```
eurekaffeine, line 102: `ContentResolver#insert()` might return nullable `mediaContentUri`, shall we check it instead of assertion`mediaContentUri!!`?

magom001: I am not doing ANY manual code changes. The goal is to convert to kotlin with minimum changes and refactor/add null checks in follow-up PRs. That is, if this PR is reviewed, accepted and merged...

magom001:
@Arjan-Zuidema hi! Could you please see if this is something maintainers will be comfortable with. If not, I will close the PR. Thanks!

android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.kt
eurekaffeine: Well noted. I think it would be better if we can get the PR merged firstly, and then, do the null-safety check for the whole project.

Arjan-Zuidema:
Hi @magom001!

I am personally not against an automatic conversion. As long it's keeping the code readable.

I will test this soon and when it's all works I can merge it. And the we can also merge the NullPointer fix from @eurekaffeine

PR #172, opened 2020-05-08
Bugfix: Error Domain=PHPhotosErrorDomain Code=-1

ShotSkydiver:
# Summary

This is a fix for the iOS-related issue when trying to save an image, as detailed here: #157 

This fix is courtesy of the code posted here: #159 

### What's required for testing (prerequisites)?
iOS 13 device
### What are the steps to reproduce (after prerequisites)?
Attempt to save an image
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    N/A     |

## Checklist

- [X] I have tested this on a device and a simulator
- [ ] I added the documentation in `README.md`
- [ ] I updated the typed files (TS and Flow)
- [ ] I added a sample use of the API in the example project (`example/App.js`)

PR #480, opened 2023-03-20
Update the way photo type data is saved

soonbee:
# Summary

On iOS, I found that the image saved through `CameraRoll.save` has a different size(byte) than the original image.
The difference in size varies depending on the image, and may increase or decrease.
In fact, different size may be an insignificant problem depending on the point of view.
But, I want to say that **The result of** `CameraRoll.save` **is not a copy of the original image.**

It is suggested to temporarily save the file in the `NSTemporaryDirectory` and then copy it to the photo library.
If you have any other comments on my suggestion or the code I wrote, you are welcome.

### before
before

### after
after

```bash
npx react-native init AwesomeProject
cd AwesomeProject
yarn add @react-native-camera-roll/camera-roll
cd ios && pod install && cd ..
```

Add below to `Info.plist`
```
<key>NSPhotoLibraryUsageDescription</key>
<string>NSPhotoLibraryUsageDescription</string>
<key>NSPhotoLibraryAddUsageDescription</key>
<string>NSPhotoLibraryAddUsageDescription</string>
```

`App.jsx` for test
```javascript
import React, {useState} from 'react';
import {Button, Image, SafeAreaView, StyleSheet, Text} from 'react-native';
import {CameraRoll} from '@react-native-camera-roll/camera-roll';

const url  = '...'; // <-- edit here

function App() {
  const [originalSize, setOriginalSize] = useState('');
  const [copiedSize, setCopiedSize] = useState('');
  const [uri, setUri] = useState(null);

  const onPress = async () => {
    setUri(url);

    const original = await fetch(url).then(res => res.blob());
    setOriginalSize(original.size);

    const type = 'photo';
    const phUrl = await CameraRoll.save(url, {type});
    const result = await CameraRoll.iosGetImageDataById(phUrl, true);
    setCopiedSize(result.node.image.fileSize);
  };
  return (
    <SafeAreaView>
      <Button title="click" onPress={onPress} />
      <Text>original size: {originalSize}</Text>
      <Text>copied size: {copiedSize}</Text>
      <Image style={styles.image} source={{uri}} />
    </SafeAreaView>
  );
}

const styles = StyleSheet.create({
  image: {
    width: 200,
    height: 200,
  },
});

export default App;
```

Edit url as below.

* http://xxx.png
* file://xxx.gif
* data:image/jpeg;base64,/9j/4AAQSk...

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### 1) Save with URL
```objective-c
assetRequest = [PHAssetChangeRequest creationRequestForAssetFromImageAtFileURL:inputURI];
```
refer to #172

### 2) Save with NSData
It was confirmed that the size of the `data` and the size of the original image were the same.
```objective-c
NSData *data = [NSData dataWithContentsOfURL:inputURI];
NSLog(@"%d", data.length); // equal to original image's size
```
So, I tried to save `data` directly to the photo library without converting it to `UIImage`.
```objective-c
NSData *data = [NSData dataWithContentsOfURL:inputURI];
PHAssetCreationRequest *request = [PHAssetCreationRequest creationRequestForAsset];
[request addResourceWithType:PHAssetResourceTypePhoto data:data options:NULL];
assetRequest = request;
```
But, the size was still not the same.

### 3) Save the file without extensions
```objective-c
NSString *fileName = [[NSUUID UUID] UUIDString];
NSURL *tmpFileURI = [NSURL fileURLWithPath:[NSTemporaryDirectory() stringByAppendingPathComponent:fileName]];
assetRequest = [PHAssetChangeRequest creationRequestForAssetFromImageAtFileURL:tmpFileURI];
```

The following error occurs.
```
Error: The operation couldn’t be completed. (Cocoa error -1.)
```

### DIFF
--- a/ios/RNCCameraRoll.mm
+++ b/ios/RNCCameraRoll.mm
@@ -149,9 +149,18 @@ RCT_EXPORT_METHOD(saveToCameraRoll:(NSURLRequest *)request
   // URLs, `data:` URIs, etc. Video URLs are passed directly through for now; it may be nice to support
   // more ways of loading videos in the future.
   __block NSURL *inputURI = nil;
+  __block NSURL *tmpFileURI = nil;
   __block PHFetchResult *photosAsset;
   __block PHAssetCollection *collection;
   __block PHObjectPlaceholder *placeholder;
+
+  void (^removeTmpFile)(void) = ^void() {
+    NSFileManager *fileManager = [NSFileManager defaultManager];
+    NSString *path = tmpFileURI.path;
+    if ([fileManager fileExistsAtPath:path]) {
+      [fileManager removeItemAtPath:path error:nil];
+    }
+  };
 
   void (^saveBlock)(void) = ^void() {
     // performChanges and the completionHandler are called on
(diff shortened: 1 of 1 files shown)


vforvasile:
Out of curiosity, what happens if you set 
const result = await CameraRoll.iosGetImageDataById(phUrl, **false**);

Maybe it tries to convert the asset when you set true.

soonbee:
Sorry for the late reply.
I've been so busy lately that it's going to take some time to fix the code.
I close this PR and re-open it when I revise it again later.

PR #492, opened 2023-05-12
fix: make useCameraRoll functions to be shallow equal

### LINKED ISSUE #485: useCameraRoll()'s getPhotos results in infinite loop
I need to display user's photos when a component is loaded:

```
   const [photos, getPhotos] = useCameraRoll();
    useEffect(() => {
        getPhotos();
    }, [getPhotos]);
```

However, this results in component to render forever in a loop (perhaps `getPhotos` object itself changes?). If I remove `getPhotos` from `useEffect`'s dependency list, it works, but then I get an eslint warning:

```
React Hook useEffect has a missing dependency: 'getPhotos'. Either include it or remove the dependency array.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)```

If I remove `getPhotos();` out of `useEffect` completely and just call it in the component body it again starts rendering in a forever loop, I doubt it's a good pattern anyway.

How can I use `getPhotos` from the hook, while not breaking the rules of hook dependencies?

santhoshvai:
# Summary

fixes #485

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅   |
| Android |    ✅     |

## Checklist

### DIFF
--- a/src/useCameraRoll.ts
+++ b/src/useCameraRoll.ts
@@ -1,4 +1,4 @@
-import {useState} from 'react';
+import {useRef, useState} from 'react';
 import type {
   GetPhotosParams,
   PhotoIdentifiersPage,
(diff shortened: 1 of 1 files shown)


Arjan-Zuidema:
@santhoshvai I assume you tested this (in simulator or device)? You did not check the box. Code is looking fine, will test this later on the day!

PR #501, opened 2023-06-22
Fix Native Arch codgen issue in RN 0.72

johnf:
# Summary

This is related to https://github.com/react-native-cameraroll/react-native-cameraroll/issues/498

It seems when you turn on new arch that codegen isn't able to compile `import type`

This fix isn't amazing as it's duplicating the type. But there already looks to be some duplication between the files.

## Test Plan

It now compiles fine, it's a fairly simple change.

Standard camera roll setup.

Compile your app

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [] I added the documentation in `README.md`
- [] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/src/NativeCameraRollModule.ts
+++ b/src/NativeCameraRollModule.ts
@@ -3,12 +3,20 @@
 // and we want to stay compatible with those
 import {TurboModuleRegistry, TurboModule} from 'react-native';
 
-import type {SubTypes} from './CameraRoll';
-
 type Album = {
   title: string;
   count: number;
 };
+
+type SubTypes =
+  | 'PhotoPanorama'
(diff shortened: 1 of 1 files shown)


johnf:
User error on my part..
Make sure you don't have `react-native-codgen` as a dependency

johnf:
After digging deeper, I realised newer codegen when enabling newArch in 0.72 still has this issue. So re-opening

Arjan-Zuidema:
I also updated the FabricExample to include react-native 72.1. So this can be tested in the example app

PR #524, opened 2023-08-24
group_name return array of albums

zatteo:
# Summary

When using a simple call to `getPhotos` like 

```
await CameraRoll.getPhotos({
    first: 100,
    include: ['filename', 'fileExtension']
  })
```

node.group_name is different between iOS and Android :

    Android : the name of the folder which can be seen as the name of the album
    iOS : 'All photos' every time

I want to get for each photos returned by getPhotos the name of every albums containing the photos. So it would be :

    Android : same value **in an array**
    iOS : an array like ['Holiday in XX', 'Party at ZZ', ...] and not 'All photos'

It is a breaking change because it now returns an array of string instead of a string.

See https://github.com/react-native-cameraroll/react-native-cameraroll/issues/513

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] I have tested this on a simulator


ios/RNCCameraRoll.mm
```diff
@@ -435,8 +437,8 @@ static void RCTResolvePromise(RCTPromiseResolveBlock resolve,
       [assets addObject:@{
         @"node": @{
           @"type": assetMediaTypeLabel, // TODO: switch to mimeType?
-          @"subTypes":assetMediaSubtypesLabel,
-          @"group_name": currentCollectionName,
```
Arjan-Zuidema, line 439: Looks like `currentCollectionName` is not used anymore, can you clean that up?

Arjan-Zuidema (changes requested):
Code is looking good, just left a comment. Would you look into that?

ios/RNCCameraRoll.mm
zatteo: Sure! Done!

Arjan-Zuidema (approved):

zatteo:
@Arjan-Zuidema Do you need or wait something for this one?

Arjan-Zuidema:
Currently working on testing it 🙂

Arjan-Zuidema:
@zatteo It looks like the performance went down drastically. From ~75ms to  ~600-700ms for getting 1000 photos (tested on an iPhone 13 Pro). Maybe we should put this feature behind an include flag now. What are your thoughts about this?

ios/RNCCameraRoll.mm
```diff
@@ -715,6 +714,18 @@ static void RCTResolvePromise(RCTPromiseResolveBlock resolve,
     return mediaSubTypeLabels;
 }
 
+- (NSArray<NSString *> *) getAlbumsForAsset:(PHAsset *)asset {
+    NSMutableArray<NSString *> *albumTitles = [NSMutableArray array];
+
+    PHFetchResult<PHAssetCollection *> *collections = [PHAssetCollection fetchAssetCollectionsContainingAsset:asset withType:PHAssetCollectionTypeAlbum options:nil];
```
Arjan-Zuidema, line 720: It looks like this is very slow. Maybe this feature should be off by default?

Arjan-Zuidema (changes requested):

ios/RNCCameraRoll.mm
zatteo: Agreed. I see no way to speed up this. So behind an include flag is a good idea. Will work on it.

zatteo: Updated! @Arjan-Zuidema What do you think? In this version flag only change something in iOS because Android is not impacted.

Arjan-Zuidema: Thanks! But I'd rather keep the api's the same for Android and iOS if possible to minimize confusion. Can you also make it to not return the albums on Android when the include flag is not set?

zatteo: *2 ~days~ months later...* I updated the code to return albums only if the include flag is set @Arjan-Zuidema

PR #526, opened 2023-08-28
Include subtype in get albums. iOS only.

zatteo:
# Summary

Return subtype in get albums to be able to check for example if an album is an iCloud Shared Albums. Combined with https://github.com/react-native-cameraroll/react-native-cameraroll/pull/525, it allows to do thing without anything related to iCloud Shared Albums, or with iCloud Shared Albums.

Before these two PRs, getAlbums returned iCloud Shared Albums, and getPhotos did not return photos in iCloud Shared Albums.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌ (not relevant)  |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,7 @@ Returns a Promise with a list of albums
 Array of `Album` object
   * title: {string}
   * count: {number}
+  * subtype: {string | undefined} : See AlbumSubType type for possible values. iOS only.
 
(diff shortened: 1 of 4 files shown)


zatteo:
Thank you!

PR #536, opened 2023-09-21
added getPhotoThumbnail

jp-23:
# Summary
Added getPhotoThumbnail method for iOS only.  These changes are based off of the ideas from @dokkis in #168.

I would like to add for Android at some point.  (I think @dokkis came close to a working solution for Android?)

Comments/Questions

1. Thumbnails are generated for both images & videos.  I've added sample code to the README.
2. I ultimately used `getPhotoByInternalID` as a starting point / guide.
3. I use `UIImageJPEGRepresentation` (returns the image as JPEG) for the base64 encoding - regardless of photo/video or extension.
4. I tried implementing caching with `PHCachingImageManager` but ran into serious performance issues once testing on my device.  When testing on the simulator I did not see performance issues - and I also did not really notice any performance increases - so I'm not entirely sure I was implementing it correctly.  I did not include those changes in this pull request - i can add them to a different branch or pull request if anyone else wants to take a look at those changes.  We could also consider some other type of caching maybe directly in JavaScript / React Native?
5. I am not sure what is meant by `flow` in the checklist below?

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,7 @@ async function savePicture() {
 * [`deletePhotos`](#deletephotos)
 * [`iosGetImageDataById`](#iosgetimagedatabyid)
 * [`useCameraRoll`](#usecameraroll)
+* [`getPhotoThumbnail`](#getphotothumbnail) **iOS only**
 
 ---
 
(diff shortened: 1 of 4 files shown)


Arjan-Zuidema (approved):
Looks good for a first version! Maybe we can iterate on this and make it also available for Android.

Issue #568, opened 2023-12-12
feature request: Return full `PhotoIdentifier` in `save()` 


mrousavy:
Hey all!

Thanks for this amazing package, RN CameraRoll is great!

I was wondering if it'd be possible for RN CameraRoll to return the full `PhotoIdentifier` information instead of just the URI on `save()`, as this is quite relevant in my case since I immediately display images after capturing them.

I'd love to help with implementation if needed!

Thanks!

Arjan-Zuidema:
Hi @mrousavy!
That sounds like a functional feature. I'd be willing to accept a pull request if you have the time to create it. Currently I don't have much time to create a feature

mrousavy:
Awesome, working on it as we speak!

mrousavy:
PR up: https://github.com/react-native-cameraroll/react-native-cameraroll/pull/569

bobbyecho:
up up up, i need this feature so much 🙏

PR #569, opened 2023-12-13
feat: Return full `PhotoIdentifier` in save()

mrousavy:
# Summary

**TL;DR: Adds `saveAsset(): Promise<PhotoIdentifier>` to get the full photo asset immediately after saving.**

---

This PR changes the behaviour of the native `saveToCameraRoll` function to return the full `PhotoIdentifier` instead of just the URI.

The motivation behind this is feature request #568.

This is useful if you want to append the image to state and render it immediately, e.g. in Gallery or Camera apps (such as [ShadowLens](https://mrousavy.com/projects/shadowlens)), because you can use the `id` for equality comparisons and keys, and elements like width and height are immediately available.

To not make this breaking, `save()` will still return a string (the URI), and a new method called `saveAsset()` has been introduced, which returns the full photo asset.

## Test Plan

Save a photo by calling `save()` and `saveAsset()` and compare the different return values. one is a string, one the full PhotoIdentifier.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -13,7 +13,7 @@ import {NativeModules} from 'react-native';
 
 // Mock the RNCNetInfo native module to allow us to unit test the JavaScript code
 NativeModules.CameraRollManager = {
-  saveToCameraRoll: jest.fn().mockReturnValue({ node: { image: { uri: 'mock://asset.jpg' } } }),
+  saveToCameraRoll: jest.fn().mockResolvedValue({ node: { image: { uri: '' } } }),
   getPhotos: jest.fn()
 };
 
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -13,7 +13,7 @@ import {NativeModules} from 'react-native';
 
 // Mock the RNCNetInfo native module to allow us to unit test the JavaScript code
 NativeModules.CameraRollManager = {
-  saveToCameraRoll: jest.fn(),
+  saveToCameraRoll: jest.fn().mockReturnValue({ node: { image: { uri: 'mock://asset.jpg' } } }),
   getPhotos: jest.fn()
 };
 
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -13,7 +13,7 @@ import {NativeModules} from 'react-native';
 
 // Mock the RNCNetInfo native module to allow us to unit test the JavaScript code
 NativeModules.CameraRollManager = {
-  saveToCameraRoll: jest.fn().mockResolvedValue({ node: { image: { uri: '' } } }),
+  saveToCameraRoll: () => Promise.resolve(({ node: { image: { uri: '' } } })),
   getPhotos: jest.fn()
 };
 
--- a/src/CameraRoll.ts
+++ b/src/CameraRoll.ts
@@ -231,6 +231,7 @@ export class CameraRoll {
   /**
    * Saves the photo or video to the camera roll or photo library, and returns the URI of the newly created asset.
    *
+   * @deprecated `save(...)` is deprected - use `saveAsset(...)` instead.
    */
   static async save(
     tag: string,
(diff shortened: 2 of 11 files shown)


Arjan-Zuidema:
@mrousavy I dont rely on the unit tests (because they did not work when I became a maintainer), I usually run tests on the FabricExample to make sure everything works.

mrousavy:
Yep I think I'm done - I tested iOS and Android and both the old `save` and new `saveAsset` funcs work! :) 

I do think that the codebase is a bit tricky to work with, maybe we could tackle a Swift/Kotlin rewrite sometime in the next year and separate the components a bit cleaner. I could help with that

santhoshvai:
That would be great

jest.setup.js
```diff
@@ -13,7 +13,7 @@ import {NativeModules} from 'react-native';
 
 // Mock the RNCNetInfo native module to allow us to unit test the JavaScript code
 NativeModules.CameraRollManager = {
-  saveToCameraRoll: jest.fn(),
+  saveToCameraRoll: jest.fn().mockResolvedValue({ node: { image: { uri: '' } } }),
```
idrisssakhi, line 16: i would say () => Promise.resolve(({ node: { image: { uri: '' } } })), so it can work

src/CameraRoll.ts
```diff
@@ -229,13 +229,25 @@ export class CameraRoll {
   }
 
   /**
-   * Saves the photo or video to the camera roll or photo library.
+   * Saves the photo or video to the camera roll or photo library, and returns the URI of the newly created asset.
    *
    */
-  static save(
+  static async save(
```
idrisssakhi, line 235: i would put a deprecation comment here so we can remove it in the future

src/CameraRoll.ts
```diff
@@ -229,13 +229,25 @@ export class CameraRoll {
...
   ): Promise<string> {
+    const asset = await this.saveAsset(tag, options);
+    return asset.node.image.uri;
+  }
+
+  /**
+   * Saves the photo or video to the camera roll or photo library, and returns the newly created asset.
+   *
+   */
+  static saveAsset(
```
idrisssakhi, line 247: please add the jsDoc params

src/__mocks__/nativeInterface.ts
```diff
@@ -2,6 +2,6 @@
 
 module.exports = {
   deletePhotos: jest.fn(),
-  saveToCameraRoll: jest.fn(),
+  saveToCameraRoll: jest.fn().mockResolvedValue({node: {image: {uri: ''}}}),
```
idrisssakhi, line 5: save feedback as previous one for the mocking

src/__tests__/CameraRollTest.ts
```diff
@@ -8,7 +8,9 @@ let mockGetPhotos: jest.Mock;
 
 jest.mock('../NativeCameraRollModule', () => {
   mockDeletePhotos = jest.fn();
-  mockSaveToCameraRoll = jest.fn();
+  mockSaveToCameraRoll = jest
+    .fn()
+    .mockResolvedValue({node: {image: {uri: ''}}});
```
idrisssakhi, line 13: same

idrisssakhi (changes requested):

idrisssakhi:
i think we should also update the PRs policy to only have one commit per PR eventhough we have the squash on github i find it better to have only one commit

idrisssakhi:
@mrousavy seems ok after testing, if you take the feedbacks into account we can merge it after that

mrousavy:
Done, applied your suggestions! :)

mrousavy:
yuhuu, thanks! let me know if there's any issues/regressions coming up :)

Arjan-Zuidema:
@idrisssakhi seems like the ci failed and no release was created

PR #585, opened 2024-02-08
fix: Fix saving assets returning `string` below Android R

mrousavy:
# Summary

This is a follow-up PR of my previous PR; https://github.com/react-native-cameraroll/react-native-cameraroll/pull/569

I forgot to add the branch for versions below Android Q, so on API 29 or lower, the `saveAsset` method returned a `string` instead of an `Asset`/`PhotoIdentifier`, causing `save` to fail.

This PR fixes this by also getting the full photo info on Android 28 or lower. 👍 

## Test Plan

Tested it in my app and it worked. Just try `save` on a device with API 28 or lower.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |           |
| Android |    ✅     |

## Checklist


mrousavy:
cc @Arjan-Zuidema @santhoshvai @idrisssakhi @bartolkaruza - sorry for forgetting that in my previous PR!

mrousavy:
Thank you!

PR #588, opened 2024-02-20
Return sourceType and fix missing filename and fileSize

zatteo:
# Summary

This PR is divided in 2 parts.

1. Return sourceType

Allow to return the source type of an asset. On iOS, the source type of an asset can be the user library (`sourceType` will be `UserLibrary`) or an iCloud Shared Album (`sourceType` will be `CloudShared`). Linked to this PR https://github.com/react-native-cameraroll/react-native-cameraroll/pull/525 which allows to return assets from iCloud Shared Albums.

It is not very interesting on Android so it will always return `UserLibrary`. 

2. and fix missing filename and fileSize

In the meantime, I fixed missing filename and fileSize as reported in this issue https://github.com/react-native-cameraroll/react-native-cameraroll/issues/576. More information in commits messages.

- Have assets in the user library and in an iCloud Shared Albums
- Check that sourceType is correctly set if in `include`
- Check that sourceType is `null` otherwise

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/ios/RNCCameraRoll.mm
+++ b/ios/RNCCameraRoll.mm
@@ -218,7 +218,8 @@ RCT_EXPORT_METHOD(saveToCameraRoll:(NSURLRequest *)request
                                                  includeImageSize:YES
                                                   includeFileSize:YES
                                           includePlayableDuration:YES
-                                                  includeLocation:YES];
+                                                  includeLocation:YES
+                                                  includeSourceType:YES];
         resolve(dictionary);
       } else {
         reject(kErrorUnableToSave, nil, error);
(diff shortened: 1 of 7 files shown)


ios/RNCCameraRoll.mm
```diff
@@ -472,18 +479,13 @@ - (NSDictionary*) convertAssetToDictionary:(PHAsset*)asset
...
       PHAssetResource *_Nullable resource = NULL;
-      NSNumber* fileSize = [NSNumber numberWithInt:0];
 
-      if (includeFilename || includeFileSize || [mimeTypes count] > 0) {
+      if ([mimeTypes count] > 0) {
         // Get underlying resources of an asset - this includes files as well as details about edited PHAssets
-        // This is required for the filename and mimeType filtering
+        // This is required for mimeType filtering
         NSArray<PHAssetResource *> *const assetResources = [PHAssetResource assetResourcesForAsset:asset];
         resource = [assetResources firstObject];
```
zatteo, line 488: These lines are now duplicated because we need `resource` here for mimeType filtering and in  convertAssetToDictionary. I do no know if it is the best way to do. Happy to hear feedback here.

Arjan-Zuidema (approved):

Arjan-Zuidema:
Code is looking fine, will test soon ™

PR #595, opened 2024-03-08
Add a new event called onProgressUpdate

mauriciomeirelles:
# Summary

Add a new event called `onProgressUpdate` that notifies the app about the progress of the image/video being downloaded from iCloud.

Sometimes when calling `iosGetImageDataById`, the image/video can be downloaded from iCloud. To be able to receive the progress of this download, you need to add a listener to the `onProgressUpdate` event and use it to render on the UI.

The event generated will be an object containing the image id (`id`) and the progress of the download (`progress`). The `id` is a string with the `internalID` you used to call `iosGetImageDataById`. The `progress` is a double ranging from 0 to 1, where 0 represents the start of the download and 1 represents the completion of the download.

```javascript

import { progressUpdateEventEmitter } from '@react-native-camera-roll/camera-roll';

useEffect(() => {
  const subscription = progressUpdateEventEmitter.addListener(
    'onProgressUpdate',
    event => {
      // Render the progress of the image / video being 
      // downloaded using event.id and event.progress
    },
  );

  return () => {
    subscription.remove();
  };
}, []);

```

Here is a video of how we are using it at Shopify Mobile app.

https://github.com/react-native-cameraroll/react-native-cameraroll/assets/5813840/00fc32c7-7b27-4049-b7fa-8811b27e5d30

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/FabricExample/ios/Podfile.lock
+++ b/FabricExample/ios/Podfile.lock
@@ -931,7 +931,7 @@ PODS:
   - React-jsinspector (0.72.1)
   - React-logger (0.72.1):
     - glog
-  - react-native-cameraroll (7.4.1):
+  - react-native-cameraroll (7.4.3):
     - hermes-engine
     - RCT-Folly (= 2021.07.22.00)
     - RCTRequired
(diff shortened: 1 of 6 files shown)


santhoshvai:
awesome, thanks for adding documentation in the README as well

Issue #615, opened 2024-04-24
Error on saveAsset (v7.6.0) [iOS]


gezquinndesign:
We're getting the following error on iOS when upgrading from 7.4.0 to the latest version (7.6.0):

<details>
<summary>Error</summary>

```js
{
  "code": "E_UNABLE_TO_SAVE",
  "message": "The operation couldn’t be completed. (PHPhotosErrorDomain error -1.)",
  "domain": "PHPhotosErrorDomain",
  "userInfo": {},
  "nativeStackIOS": [
    "0   Elsewhere Beta                      0x0000000100c48be8 RCTJSErrorFromCodeMessageAndNSError + 112",
    "1   Elsewhere Beta                      0x0000000100be56c4 __41-[RCTModuleMethod processMethodSignature]_block_invoke_2.73 + 152",
    "2   Elsewhere Beta                      0x0000000101326428 __57-[RNCCameraRoll saveToCameraRoll:options:resolve:reject:]_block_invoke.70 + 916",
    "3   Photos                              0x00000001af2d0454 72B67AD1-90CA-3201-A098-8F925C71C3B9 + 971860",
    "4   Photos                              0x00000001af2d1fdc 72B67AD1-90CA-3201-A098-8F925C71C3B9 + 978908",
    "5   Photos                              0x00000001af2d1a00 72B67AD1-90CA-3201-A098-8F925C71C3B9 + 977408",
    "6   Photos                              0x00000001af2d15f8 72B67AD1-90CA-3201-A098-8F925C71C3B9 + 976376",
    "7   CoreFoundation                      0x00000001979fbc84 3A5F992A-D1CD-312E-BD2E-F7C66343A417 + 203908",
    "8   CoreFoundation                      0x00000001979fb710 3A5F992A-D1CD-312E-BD2E-F7C66343A417 + 202512",
    "9   Foundation                          0x00000001968c6698 D92E19C1-6299-3E94-8614-C505D5ABCCDB + 415384",
    "10  Foundation                          0x00000001968c6218 D92E19C1-6299-3E94-8614-C505D5ABCCDB + 414232",
    "11  Foundation                          0x00000001968c5b80 D92E19C1-6299-3E94-8614-C505D5ABCCDB + 412544",
    "12  libxpc.dylib                        0x00000001f3a0dff8 042EA33B-3C8F-3FB8-869C-369A3115977C + 126968",
[...]

PR #616, opened 2024-04-24
fix(ios): Revert back to creationRequestForAssetFromImage for normal …

gezquinndesign:
# Summary

In commit 73f34f6, norml image processing in the `save`/`saveAsset` function was changed from `creationRequestForAssetFromImage` to `creationRequestForAssetFromImage` to `creationRequestForAssetFromImageAtFileURL`, this led to issue #615. This PR reverts that code to the previous implementation.

N/A

Pass any valid file uri to`saveAsset` on ios, eg:

```js
CameraRoll.saveAsset(remoteAssetLibraryURL, { type: 'photo' })` 
```

and it should no longer result in an error.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅   |
| Android |    N/A     |

## Checklist

### DIFF
--- a/ios/RNCCameraRoll.mm
+++ b/ios/RNCCameraRoll.mm
@@ -196,7 +196,9 @@ RCT_EXPORT_METHOD(saveToCameraRoll:(NSURLRequest *)request
           assetRequest = [PHAssetChangeRequest creationRequestForAssetFromImage:webpImage];
         } else {
           // normal Image (jpg, heif, png, ...)
-          assetRequest = [PHAssetChangeRequest creationRequestForAssetFromImageAtFileURL:inputURI];
+          NSData *data = [NSData dataWithContentsOfURL:inputURI];
+          UIImage *normalImage = [UIImage imageWithData:data];
+          assetRequest = [PHAssetChangeRequest creationRequestForAssetFromImage:normalImage];
         }
       }
       placeholder = [assetRequest placeholderForCreatedAsset];


Arjan-Zuidema:
After looking at the history of the file I seem to have found why it was changed. Looks like it was changed to copy exif when saving: https://github.com/react-native-cameraroll/react-native-cameraroll/commit/73f34f6fa2f7075836a0794d7a2ccfaeb3c47b78

@mrousavy do you have any thoughts on this?

Reverting this could make exif not work again, can you confirm that exif is still being saved @gezquinndesign

mrousavy:
Hey - yea well I wrote that code to support EXIF data.
This is also the same method that is used for video files, so I don't think it makes sense to use a different approach for photos just so the caller doesn't have to manually download the photo first.

Again, you'd need to do the same for video, and there's no alternative as there is for photo.

So I think the best approach is to download first and then save.
That way the proper metadata is transferred (any tags, GPS location, lens info, etc etc).

mrousavy:
I use the GPS EXIF metadata transfer method in production for ShadowLens, and it works great.

gezquinndesign:
@mrousavy Sorry, I noticed late that the commit was actually part of a focused PR.  I've found the [real cause](https://github.com/react-native-cameraroll/react-native-cameraroll/issues/615#issuecomment-2079647254) of my issue and followed the approach suggested by you and @Arjan-Zuidema.

Sorry for leading everyone on a wild goose chase and thanks for all your help.

PR #601, opened 2024-03-22
Added noop add/remove listener functions for Android

Arjan-Zuidema:
# Summary

Added missing listeners for Android in Fabric, this fixes #600 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

- [x] I have tested this on a device and a simulator
- [ ] ~I added the documentation in `README.md`~
- [ ] ~I updated the typed files (TS and Flow)~
- [ ] ~I added a sample use of the API in the example project (`example/App.js`)~

PR #619, opened 2024-04-30
Fix add removeListeners argument type

johnf:
# Summary

New arch support was fixed in #601 but the type wasn't quite correct

## Test Plan

COmpilation of my app now succeeds

RN0.74 with newarch enabled

It willcompile without errors

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |   N/A     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
+++ b/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java
@@ -1090,5 +1090,5 @@ public class CameraRollModule extends NativeCameraRollModuleSpec {
   }
 
   public void addListener(String eventName) {}
-  public void removeListeners(double count) {}
+  public void removeListeners(int count) {}
 }


kesha-antonov:
`int` should be `double`

https://github.com/react-native-cameraroll/react-native-cameraroll/blob/master/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java#L1101

https://github.com/facebook/react-native/blob/93c079b92a028354cb7926b2029601238bf82d7d/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java#L421

Otherwise it gives error when building

fatemekarami:
> `int` should be `double`
> 
> https://github.com/react-native-cameraroll/react-native-cameraroll/blob/master/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java#L1101
> 
> https://github.com/facebook/react-native/blob/93c079b92a028354cb7926b2029601238bf82d7d/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/modules/websocket/WebSocketModule.java#L421
> 
> Otherwise it gives error when building
> 
> `node_modules/@react-native-camera-roll/camera-roll/android/src/main/java/com/reactnativecommunity/cameraroll/CameraRollModule.java:67: error: CameraRollModule is not abstract and does not override abstract method removeListeners(double) in NativeCameraRollModuleSpec public class CameraRollModule extends NativeCameraRollModuleSpec {`


  convert public void removeListeners(int count) {} to public void removeListeners(double count) {} worked for me in CameraRollModule.java file

johnf:
@Arjan-Zuidema Can you roll back this commit. I don't know why this was working for me before but it's broken now.
I think I must have had something weird cached.
Sorry for screwing this up

Arjan-Zuidema:
> @Arjan-Zuidema Can you roll back this commit. I don't know why this was working for me before but it's broken now. I think I must have had something weird cached. Sorry for screwing this up

I did a roll back, was not able to test this now. But I can't remember how this did not show up in my tests. Probably need better checks on CI.

